PR #554, opened 2015-01-07
Link love

johnlucia:
@dreuss I worked on that stuff we did yesterday, and this is what I was able to beat it down to.  I still need to test that new helper method, but other that that I think we're ready to go.  Can you think of anything that we're missing?


app/views/web_templates/show.html.erb
```diff
@@ -42,6 +42,11 @@
 <% content_for(:bottom_of_body) do %> 
   <% if mode == "preview" %>
     <%= javascript_include_tag(*web_template.javascripts) %>
+
```
johnlucia: I originally tried to use the rails javascript_include_tag helper here, but since that always adds the `//<![CDATA[` thing to the tag, the json.parse call would blow up. So, I just built the tag by hand.

app/assets/javascripts/web_template/link_tweaker.coffee
```diff
@@ -0,0 +1,10 @@
+$ ->
+  previewConfigs = JSON.parse($('#preview-configs').html())
+
+  # Prepend the URN to the path of anything that looks like an internal page link
+  $('body').delegate('a', 'click', ->
+    linkHref = $( this ).attr('href')
+    if typeof linkHref != 'undefined'
+      previewHref = linkHref.replace("apartments/", previewConfigs.urn + "/apartments/")
```
dreuss: Are you going to sub out "apartments/" with the previewConfigs.slug instead of hard coding it to only look for apartments in case the industry is different?

spec/helpers/web_template_helper_spec.rb
```diff
@@ -31,4 +31,15 @@
...
   end
+
+  describe "preview_configs" do
+    let(:preview_configs) { helper.preview_configs(params, web_home_template) }
+
+    it "defines the location URN" do
+      JSON.parse(preview_configs)["urn"].should eq(params["urn"])
+    end
+    it "defines the website slug" do
+      JSON.parse(preview_configs)["slug"].should eq(web_home_template.website.client.vertical_slug)
```
dreuss: Any reason for the switch to using web_home_template in the spec vs. using web_template as it's being called by app/views/web_templates/show.html.erb.

app/assets/javascripts/web_template/link_tweaker.coffee
```diff
@@ -0,0 +1,15 @@
+$ ->
+  previewConfigs = JSON.parse($('#preview-configs').html())
+
```
johnlucia: @dreuss I figured we should probably add your check for http/https to the location stuff too. So, I rewrote that check as a regex and moved everything into the same `delegate` call, and used previewConfigs.corporate to conditionally set the previewHref var. 

I think this still all works, right?

johnlucia:
@bbauer @mmitchellg5  Can I get one or both of you guys to take a look at this?

Right now, the url structure is different in preview mode than it is on deployed sites, and that's a bit hard on everyone who deals with the CMS. This change adds some js in preview mode that takes the deployed site url structure and tweaks it to work in preview mode so we can click and follow links without getting a rails error.

@dreuss was in on this too. I'm gonna get this pushed out to a forked CMS app for some QA next. I'll add in a link to that once it's up and running.

johnlucia:
here it is in action...  https://g5-cms-googus.herokuapp.com/

vs. the production cms, where all the relative links in preview mode result in an error... https://g5-cms-1slhp2tc-compass-rock-r.herokuapp.com

ceckert:
Very cool. Is there any way to get the entire browser window to reload, so the top of the frame (the widgets on page, etc) is also shown? Meaning, if I navigate from Home to Photos, can the top frame also load the photos page? If that is too much effort, maybe a label on the nav bar to indicate the widgets in the top frame are for the Home page, even though I have just navigated the preview frame to the Photos page?

image

johnlucia:
@ceckert the preview at the bottom half of the edit screen is an iframe, so we can't get actions in that section to affect the page it's self... or at least I don't know of any way to do it

ceckert:
OK. Makes sense. Great improvement regardless!

app/helpers/web_templates_helper.rb
```diff
@@ -26,4 +26,12 @@ def head_widgets(web_template)
   def meta_description_widgets(web_template)
     web_template.meta_description_widgets.map(&:render_show_html).join
   end
+
+  def preview_configs(params, web_template)
+    { urn: params['urn'],
+      slug: web_template.website.client.vertical_slug,
+      corporate: web_template.website.corporate?,
+      slug_corporate: web_template.website.web_home_template.preview_url }.to_json.html_safe
```
variousred: this makes me nervous because we're now managing url structures in two places, in the ruby app, and now in javascript. these can easily diverge

variousred:
@johnlucia can you provide some examples of the specific issue, the differences in urls i'm not 100% clear what is trying to be fixed here

johnlucia:
@variousred here's the problem:

go here... https://g5-cms-1slhp2tc-compass-rock-r.herokuapp.com/g5-cl-1slhp2tq-vizcaya-hilltop/apartments/nv/reno/amenities ...and click on any link. 

The problem is, a page url on a deployed site looks like this: 
`/apartments/nv/reno/amenities/`

the same page in preview mode has a url that looks like this: 
`/g5-cl-1slhp2tq-vizcaya-hilltop/apartments/nv/reno/amenities`

All the links in preview mode use the deployed site link structure, so that's why you get an error anytime you click on a link. That's a significant pain in the ass for builders and QA.  

I think the whole reason behind that is the original architecture never considered a situation where one client had multiple locations in the same City/State. So, the `/g5-cl-1slhp2tq-vizcaya-hilltop` garbage was added to page paths so the app could figure out what location it was supposed to serve a page for in preview mode. I think @bbauer was in on the original fix for the multiple locations in one city problem, so he probably knows more about that.

All we're doing (in preview mode only) is looking at each link on click, and if it looks like an internal link, then prepend the `/g5-cl-1slhp2tq-vizcaya-hilltop` stuff. So, this doesn't really manage URL structure so much as it just uses javascript to make preview links work like most people would expect them to.

variousred:
@johnlucia gotcha
:shipit:

PR #555, opened 2015-01-07
Singledomain preview

variousred:
fix for showing previews in cms


variousred:
@bbauer https://www.pivotaltracker.com/story/show/85356354

app/models/web_home_template.rb
```diff
@@ -27,3 +27,4 @@ def relative_path
     "/"
   end
 end
+
```
bbauer: Has anyone really ever had to cat two files together, in the history of G5?

bbauer:
:+1:

app/models/web_home_template.rb
variousred: I LIKE THE WAY IT LOOKS IN VIM!!!!!

bbauer: FINE!

bbauer: :) Honestly I'm fine either way...

variousred: :D

PR #556, opened 2015-01-13
Column dropdown fix

levibrown:
move `update_row_widget_garden_widgets_setting` and `update_column_widget_garden_widgets_setting` to `update_all`
those methods only need to be called once, not per widget.

this will also allow page builders to use update widgets to fix the nil dropdown in the column widget in the case when widget timestamps already match.

https://www.pivotaltracker.com/n/projects/877197/stories/85928238


variousred:
:shipit:

PR #557, opened 2015-01-14
Compass Rock Corp deploy is broken.

ceckert:
Not sure why the code from @variousred would have ever worked. Per the Location.corporate scope and the tests, Location.corporate should always return a Location, not an association.


ceckert:
@bbauer can you review?

ceckert:
@bbauer not sure why there were four files in this PR. I only edited the one. Let me know if you think that will be a problem.

bbauer:
@ceckert those are VCR cassette files that get written when you run specs (if they haven't been run for a while locally) so you're good to go. Anyway, this looks good to me.

:+1:

PR #559, opened 2015-01-14
Widget seed

variousred:
there is no widget with a slug of "phone", assuming this should be phone-widget

will log widget creation errors like this in the future at the debug level.


bbauer:
:+1:

PR #560, opened 2015-01-14
fixes the issue of assets not saving when there are no categories

variousred:
https://www.pivotaltracker.com/story/show/86125774


bbauer:
Sounds legit :+1:

ceckert:
:shipit: Thanks!

PR #562, opened 2015-01-16
Revision

variousred:
Removes initializer to set revision as a constant. Removes reference to constant in the header bar. Stupid heroku deletes the .git directory before slug compilation to save space.

### DIFF
--- a/config/initializers/1git_revision.rb
+++ b/config/initializers/1git_revision.rb
@@ -1,3 +1,0 @@
-module G5CMS
-  REVISION = `git log --pretty=format:'%h' -n 1`
-end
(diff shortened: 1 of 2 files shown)


variousred:
@bbauer @ceckert

levibrown:
http://stackoverflow.com/questions/8792716/reflecting-heroku-push-version-within-the-app#answer-15367505

variousred:
```
variousred Any good tips on displaying a git sha in a heroku app besides git pre-push hooks
catsby variousred: if you know the slug id, you could query the API for that information
catsby knowing the slug id is the trick...
variousred do i need to be magical or just see through an illusion
catsby whisper magical incantations, I'm afraid
catsby ...
variousred :p
catsby other than that, I'll just say "soon"
variousred O.o
catsby It's a common request
```

variousred:
@bbauer :/

bbauer:
:+1:

PR #565, opened 2015-01-27
WIP - Smarter widgets

johnlucia:
There's quite a bit of refactoring still to be done here, but the idea is something like this...

1. Run a migration to add a parent_id column to Widgets, and populate that column for all widgets nested inside a row or column widget.

2. Add a new `location_settings` method that is available in the liquid and uses the new parent_id value to find it's way up to some key web template and location data.

### DIFF
--- a/app/decorators/setting_decorator.rb
+++ b/app/decorators/setting_decorator.rb
@@ -41,6 +41,8 @@ class SettingDecorator < Draper::Decorator
     # going this route requires we set up a placeholder setting in index.html of the widget, and
     # then call .location_settings on that placeholder within the liquid.
 
+    # also need to fine tune the migration that assigns the new parent_id attribute.
+
     parent = Widget.find_by_id owner.parent_id
     if parent
       { parent_id: owner.parent_id,
--- a/app/decorators/setting_decorator.rb
+++ b/app/decorators/setting_decorator.rb
@@ -44,7 +44,7 @@ class SettingDecorator < Draper::Decorator
     parent = Widget.find_by_id owner.parent_id
     if parent
       { parent_id: owner.parent_id,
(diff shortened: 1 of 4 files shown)


johnlucia:
@ceckert @bbauer @variousred Can you all take a look at this?  According to Eckert there was some collaboration on another way to pull Location and WebTemplate info down to child widgets. You guys have a better way?

ceckert:
@johnlucia the pull request I mentioned was here: https://github.com/G5/g5-content-management-system/pull/491

ceckert:
@johnlucia so, based on that PR and refactoring since then, this code should give you what you want. It's a little non-intuitive, because it is using the setting, not the widget. It is using a copy of the CR database.

w2 = Widget.find(7477) # A photo widget in a content stripe
wfb = WebsiteFinder::Base.new(w2)
setting = wfb.layout_setting_for(w2.id)
wfs = WebsiteFinder::Setting.new(setting).find
wfs.owner # this is the location

johnlucia:
@ceckert Unless I'm missing something, this code is still fully inaccessible from the widget garden... Right? That's the problem we are still working on. The widget garden has very limited access to any ruby code, and only then through the liquid template stuff.

ceckert:
@johnlucia You are right. The widget would still need the setting, but the value would be set in Ruby (concern or otherwise) when it is added to a page.

ceckert:
@bbauer your thoughts on this? I don't want to build too much dependency between the widgets and the CMS code, but there are numerous widgets that need the client and/or location urn.

variousred:
@ceckert @johnlucia client is easy because there always only be one client, so you dont have to traverse up a tree from a widget to get there. Location could be a bit trickier

PR #570, opened 2015-02-12
Update widget_cloner.rb

variousred:
pp not working in production, changes to rails logger debug level

@ceckert @bbauer


ceckert:
Should we just rip out these log statements all together? Deep Clone seems to be working OK. @variousred do you see value in leaving them in?

ceckert:
:+1:  :ship:

PR #575, opened 2015-02-24
Areapage

mmitchellg5:
fixes issues of scoping, no corp designation.

changes the scope to a class method so we always know what we're getting. This was causing confusion when the scope would return a single record in the case of one existing, and a relation in the case when there were no hits.

This allows for a deploy to happen in the case of no location with corp designation by substituting for an arbitrary location when certain attributes must be found.

https://www.pivotaltracker.com/story/show/89029990

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -18,11 +18,14 @@ class Location < ActiveRecord::Base
 
   default_scope { order("corporate DESC") }
 
-  scope :corporate, -> { where(corporate: true).first }
   scope :live, -> { where(status: "Live") }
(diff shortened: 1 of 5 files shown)


mmitchellg5:
@ceckert @bbauer

bbauer:
Not sure why this isn't using the scope, but may be an issue...
https://github.com/G5/g5-content-management-system/blob/dcb3dfdd4dded00df7f34822e14436bcb87228ba/lib/corporate_navigation_setting.rb#L15

mmitchellg5:
I think thats ok

bbauer:
:+1:

PR #576, opened 2015-02-24
 adds spec for excluding setting clone on analytics widget, excludes ana...

mmitchellg5:
adds spec for excluding setting clone on analytics widget, excludes analytics widgets settings from cloning

https://www.pivotaltracker.com/story/show/88949542

@ceckert


ceckert:
:+1: :shipit:

gladdeng:
@mmitchellg5 looks like tests have cleared. @ceckert are we good to go?

PR #578, opened 2015-02-26
Advanced simple spinup

thescubageek:
Expands simple spinup to be far more robust... enhancements include:

1. Refactors WebsiteSeeder into Seeder module, which enables far more granular Seeding of individual widgets, nested layout widgets, drop targets, web pages, web themes, web layouts, and websites. 

2. Enables spinup with nested widgets inside Column and Content Stripe Widgets using the ColumnWidgetSeeder and ContentStripeWidgetSeeder, both an extension of the abstract class LayoutWidgetSeeder - finally makes it possible to truly have prefab Column and Content Stripe widgets from yaml files.

3. Adds SeederSerializers for each corresponding Seeder, which converts existing widgets, drop targets, web pages, and websites into generic yaml seeder files - these seeder yaml file can then be passed to the corresponding Seeder to create new instances from the seeder information.

4. WebsiteSeederSerializer, WebPageTemplateSeederSerializer, and ContentStripeWidgetSeederSerializer have a new .to_yaml_file() method, which exports the seeder yaml file to a directory in config/defaults and names the files according to their location / page (for websites / pages) or their widget contents (for content stripes). These files will be used as default files for creating granular spinup controls for individual pages and content stripes, and can also be used to spinup specific website controls. 

5. Adds liquid parsing to GardenWidgets using the new liquid setting, which enables Web Builders to use liquid variables in the various settings for liquid-enabled widgets. Current, only the HTML, Meta Description, and Photo widgets have liquid enabled. (See PR: https://github.com/G5/g5-widget-garden/pull/449). Parsing of liquid parameters is handled in LiquidParamters module.

6. Adds parent_widget method to widgets model, which enables child widgets to quickly reference their parent and other ancestor widgets.

7. Adds set_setting and get_setting methods to widgets model, which makes it much DRYer to access widget-specific settings.

8. Adds get_child_widget and set_child_widget methods to widgets model, which makes it much DRYer to access and modify the child widgets in layout widgets (Content Stripe and Column). Also guarantees correct selection of child widget in cases when orphan child widgets exist.

9. Adds ContentStripeWidgetSeederSerializerJob, which makes it possible to scrape all Content Stripes patterns across all the websites / locations for the current Client and save their YML Seeder files to the correct directory in /config/defaults/content_stripes/. Makes it easy to quickly build up a library of Content Stripe Seeder files - _caveat_: how to make this work on a cross-client way?

10. Migrates Settings for Content Stripe and Column widgets to use integer-based nomenclature, aka "row_1_column_name" versus "row_one_column_name" - greatly helps DRY up the code.

11. Creates Seeder jobs for Content Stripes, Web Page Templates, and Websites - uses Resque to add job for seeding objects based on Seeder YML files.

12. Adds API endpoints to GET the list of available Seeder YML files for Websites, Web Page Templates, and Content Stripes.

12. Adds API endpoints to POST 'serialize', which saves YML Seeder file for a given Websites, Web Page Templates, and Content Stripes.

13. Adds API endpoints to POST 'seed', attempts to create a new Website, Web Page Template, or Content Stripes based on the specified YML Seeder file.


lib/garden_widget_updater.rb
```diff
@@ -104,6 +105,12 @@ def get_thumbnail(component)
     end
   end
 
+  def get_liquid(component)
+    if component.respond_to?(:liquid)
+      ActiveRecord::ConnectionAdapters::Column.value_to_boolean(component.liquid.to_s)
```
variousred: 'true' will be cast as boolean in activerecord internally (i think using value_to_boolean) already, so you can just return the string

spec/lib/seeder/website_seeder_spec.rb
```diff
@@ -145,88 +132,94 @@ def setting_value_for(name)
...
-
-    context "when widget does not have setting" do
-      before  { subject }
-
-      it "does not set value from yml file" do
-        widget.settings.find_by_name("google_plus_id").should be_nil
-      end
-    end
-  end
+#  describe "#create_drop_targets" do
```
variousred: whats going on here

thescubageek: This is getting refactored into other seeder specs, it no longer belongs in the website_seeder spec but I'm keeping the code around and commented out as I pull it apart.... still WIP.

variousred: gotcha

lib/garden_widget_updater.rb
```diff
@@ -104,6 +105,12 @@ def get_thumbnail(component)
     end
   end
 
+  def get_liquid(component)
+    if component.respond_to?(:g5_liquid)
+      component.g5_liquid.to_s
```
variousred: w00t!

config/initializers/load_defaults.rb
```diff
@@ -1,4 +1,11 @@
-WEBSITE_DEFAULTS = YAML.load_file("#{Rails.root}/config/defaults.yml")
-WEBSITE_DEFAULTS_APARTMENTS = YAML.load_file("#{Rails.root}/config/website_defaults_apartments.yml")
-WEBSITE_DEFAULTS_SENIOR_LIVING = YAML.load_file("#{Rails.root}/config/website_defaults_senior_living.yml")
-WEBSITE_DEFAULTS_SELF_STORAGE = YAML.load_file("#{Rails.root}/config/website_defaults_self_storage.yml")
+WEBSITE_DEFAULTS_PATH = "#{Rails.root}/config/defaults/websites"
```
thescubageek: I'd like to come up with a more modular way of parsing this from the existing files...

app/models/widget.rb
```diff
@@ -102,8 +151,65 @@ def nested_settings
     widgets.collect {|widget| widget.settings}.flatten
   end
 
+  def liquid_parameters
+    return {} unless liquid
+    template = get_web_template
+    client = template.client
+    location = template.owner
```
ryankbales: Maybe pull these params out into it's own lib/module.

app/serializers/column_widget_seeder_serializer.rb
```diff
@@ -0,0 +1,65 @@
+class ColumnWidgetSeederSerializer < LayoutWidgetSeederSerializer
+
```
ryankbales: Had chat with @thescubageek about the possibility of creating migration to rename the string version of the numbers into actual integers to help clean up these field names

app/models/widget.rb
```diff
@@ -114,9 +220,15 @@ def widget_settings
...
-    widget_settings.map(&:value).map do |id|
-      Widget.find(id) if Widget.exists?(id)
-    end
+  def parent_setting
+    Setting.where("value = ?", id.to_yaml).find do |setting|
+      setting.name =~ /(?=(column|row))(?=.*widget_id).*/
+    end unless drop_target
+  end
+
+  def position_name(index)
```
thescubageek: I'd like to refactor this in the next PR for the CMS... writing a migration to change all row_one_widget_name (etc) to row_1_widget_name will make a lot of this garbage unnecessary.

app/serializers/column_widget_seeder_serializer.rb
```diff
@@ -0,0 +1,65 @@
...
+  def row_count
+    object.get_setting('row_count').try(:value)
+  end
+
+  def widgets
+    nested_widgets(widget_list)
+  end
+
+  def widget_list
```
thescubageek: All this row_one_widget_id stuff can be DRYed up if we migrate the settings name to be row_1_widget_id (etc)... will eliminate a lot of this garbage.

app/serializers/content_stripe_widget_seeder_serializer.rb
```diff
@@ -0,0 +1,77 @@
...
+  def row_layout
+    object.get_setting('row_layout').try(:value)
+  end
+
+  def widgets
+    nested_widgets(widget_list)
+  end
+
+  def widget_list
+    list = ['column_one_widget_id']
```
thescubageek: Same comment as on the ColumnWidgetSeederSerializer... need refactor of Widget and migration of settings names in next PR?

app/models/widget.rb
thescubageek: This has been moved into its own Mixin module which is used by both the WebTemplate and Widget models... enables us to control all liquid variables from a single place: https://www.pivotaltracker.com/n/projects/877197

app/serializers/column_widget_seeder_serializer.rb
thescubageek: This has been done. Row and column settings now use integers instead of string names for the numbers (1 vs one)

app/jobs/web_page_template_seeder_serializer_job.rb
```diff
@@ -0,0 +1,14 @@
+class WebPageTemplateSeederSerializerJob
+  extend HerokuResqueAutoscaler if Rails.env.production?
+  @queue = :seeder_serializer
+
+  def self.perform
+    Client.first.locations.each do |l|
```
bbauer: I'm nitpicking, but in general, I'd prefer location vs l here...and everywhere else

variousred: with bbauer, also no need to reach through Client to get at locations right? Location.all

thescubageek: Makes sense

app/models/garden_widget.rb
```diff
@@ -13,6 +13,7 @@ class GardenWidget < ActiveRecord::Base
   validates :widget_id, presence: true
   validates :slug, presence: true
   validates :url, presence: true
+  validates :liquid, presence: true, allow_blank: true
```
bbauer: Why would we validate for presence but allow an empty string? Do we just not want nil? Just curious?

thescubageek: because it was trainwrecking a bunch of specs that didn't have liquid in their yml files... i don't think it's necessarily the right solution, but it let me get past that blocker... would welcome a better solution

app/serializers/drop_target_seeder_serializer.rb
```diff
@@ -0,0 +1,11 @@
+class DropTargetSeederSerializer < ActiveModel::Serializer
+  attributes  :html_id,
+              :widgets
+
+  def widgets
+    object.widgets.inject([]) do |arr, w|
```
bbauer: You could just use map here and you wouldn't have to return arr...

```Ruby
object.widgets.map do |widget|
  WidgetSeederSerializer.new(widget, {root: false}).as_json
end
```

app/models/garden_widget.rb
bbauer: Cool :+1:

bbauer:
Steve, this code looks great. Nice work! :ice_cream: :art:

app/serializers/drop_target_seeder_serializer.rb
bbauer: You should also look into the each_with_object method in certain applications

thescubageek: I've refactored this one and a few others into map

In a lot of cases I'm not wanting to add anything to the array unless it passes a certain conditional... hence the bulky injects. But it seems I can achieve the same in a map if I follow it up with a compact to trim out the nils.

PR #579, opened 2015-02-27
adds rewrite directives to htaccess when website owner has secure_domain

variousred:
https://www.pivotaltracker.com/story/show/89246028


variousred:
@levibrown @bbauer

variousred:
@riverblufftrail this may be overkill for the story at hand. The only requirement was that links from https dont get redirected to http, correct? 
This implementation will also have the effect of ENSURING https at all times. Maybe thats not a requirement, maybe thats not needed.
This was the fastest solution which was a requirement at the time, but seems to have been allayed. 
your thoughts?

ceckert:
@variousred ensuring https at all times is fine. :+1: :shipit:

PR #580, opened 2015-02-27
Fix deploys for Locations that have orphaned websites.

ceckert:
@variousred please review.

### DIFF
--- a/lib/client_deployer.rb
+++ b/lib/client_deployer.rb
@@ -31,7 +31,7 @@ module ClientDeployer
 
   def self.compile_location_websites
     LOGGERS.each {|logger| logger.debug("compile_location_websites")}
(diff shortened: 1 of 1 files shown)


variousred:
@ceckert looks good, maybe deploy this as a branch to berkshire, ensure a-ok, then merge to master, once green

variousred:
client_deployer_spec actually has a spec that doesnt run by default — see the following comment from the spec file [https://github.com/G5/g5-content-management-system/blob/master/spec/lib/client_deployer_spec.rb]:
#
# CAUTION!!!
#
# This spec actually attempts to deploy a client location
# site. It must make outside HTTP requests to do this.  It
# does not run by default. To run this spec run:
#
# rspec spec -t deployment
#
# You will need the deployment enviroment varibales to be set.
#

PR #581, opened 2015-02-27
Fix Holland cloner issue.

ceckert:
@variousred I'm sure there is a more ruby like way to handle this. If so, please advise. This was tested on the Holland CMS and fixed the issue.


variousred:
:shipit:

PR #585, opened 2015-03-07
Fix deploy from Location page

ceckert:
See https://www.pivotaltracker.com/n/projects/877197/stories/85540646
@variousred @levibrown can I get a review? Also, please see if same issue exists in you ember-cli branch.

Smoke tested by pushing branch to this demo CMS and confirmed it works: https://g5-cms-1t9dz1wn-jm-real-estate.herokuapp.com


variousred:
:shipit:

PR #586, opened 2015-03-07
Fix page cards so edit/preview buttons show for Compass Rock corp location

ceckert:
The location uses lots and lots of content stripes. This is a work around so the buttons show on pages that use a large number of content stripes, until a better solution is available.

### DIFF
--- a/app/assets/javascripts/templates/website/_web_page_template_front.hbs
+++ b/app/assets/javascripts/templates/website/_web_page_template_front.hbs
@@ -10,6 +10,7 @@
       {{/each}}
     </ul>
       </div>
+      <div style="height: 10px">&nbsp;</div>
     <div class="buttons">
       {{#if isWebHomeTemplate}}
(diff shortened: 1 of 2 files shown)


ceckert:
@levibrown @variousred I'm not a big fan of this, but it is quick and doesn't look too bad.

image

variousred:
:shipit:

levibrown:
not a fan of this either.  really this page needs to be refactored to show the pages as row, which would be indented to show sub pages and would remove the content indicators.  
in the meantime I think we should truncate the listed icon to maintain the current card size.
we could use css or slice the array in ember:
```
li:nth-child(-n+5) {
    display: none;   
}
```

ceckert:
@levibrown, @variousred would you mind taking another look at this? I have the div height fix with no overflow. 

image

variousred:
nice! :shipit:

PR #590, opened 2015-03-13
[WIP] adds methods to widget drop for generating urls from page slugs

variousred:
this changes the way cta links are generated and stored.
Instead of storing the path as a setting which requires jobs to fire off when a location changes,
this merely stores the page slug. Paths are evaluated on the fly, and always accurate.

An added benefit from this approach is that cloning a location with ctas will just link to the right pages without running a settings update task

WIP because considerations for migrating to this approach are necessary for existing CTA widgets in the wild.

Corresponding widget garden change:
https://github.com/G5/g5-widget-garden/pull/451


app/assets/javascripts/web_template/link_tweaker.coffee
```diff
@@ -4,8 +4,9 @@ $ ->
   $('body').delegate('a', 'click', ->
     linkHref = $( this ).attr('href')
 
-    # Don't mess with anything if linkHref starts with "http" or "//"
-    if typeof linkHref != 'undefined' and !linkHref.match(/^http|^\/\//i)
+    # Don't mess with anything if linkHref starts with "http" or "//" or already begins with urn
+    pattern = "^http|^\/\/|^" + previewConfigs.urn
+    if typeof linkHref != 'undefined' and !linkHref.match(new RegExp(pattern, 'i'))
```
variousred: @johnlucia one step in removing the need for this tweaker!
CTA saves its settings as slugs and the url is built on the fly, we can call preview_url for the preview

app/controllers/web_templates_controller.rb
```diff
@@ -10,7 +10,7 @@ def show
     if @website
       if params[:web_template_slug]
         @web_template = @website.web_page_templates.where(
-          "lower(slug) = ?", params[:web_template_slug].to_s.downcase).first
+          "lower(slug) = ?", params[:web_template_slug].to_s.downcase).first || @website.web_home_template
```
variousred: this is because the routes can be a bit ambiguous
```
get "/:urn/:vertical_slug/:state_slug/:city_slug", to: "web_templates#show"
  get "/:urn/:vertical_slug/:state_slug/:city_slug/:web_template_slug", to: "web_templates#show"
  get "/:vertical_slug/:state_slug/:city_slug/:owner_urn/:web_template_slug", to: "web_templates#show"
  get "/:urn/:vertical_slug/:state_slug/:city_slug/:owner_urn/:web_template_slug", to: "web_templates#show"```

File: app/assets/javascripts/web_template/link_tweaker.coffee [hunk 3]
```diff
@@ -4,8 +4,9 @@ $ ->
   $('body').delegate('a', 'click', ->
     linkHref = $( this ).attr('href')
 
-    # Don't mess with anything if linkHref starts with "http" or "//"
-    if typeof linkHref != 'undefined' and !linkHref.match(/^http|^\/\//i)
+    # Don't mess with anything if linkHref starts with "http" or "//" or already begins with urn
+    pattern = "^http|^\/\/|^" + previewConfigs.urn
```
johnlucia: If I understand this correctly... I'm not sure if we want the `previewConfigs.urn` to be part of the regex we're looking for on the next line. The `!linkHref.match(/^http|^\/\//i)` check was to make sure we were excluding all the page's external links from our URL hackery.  I think this will correctly leave alone the urls you've already fixed, but will try to "fix" links to external pages.

variousred: @johnlucia it was `fixing` correct links, they were ending up with double urn's at the beginning

app/drops/widget_drop.rb
```diff
@@ -39,8 +39,42 @@ def corporate_navigation
...
+    pages = [WebsiteFinder::Widget.new(widget).find.web_page_templates.navigateable.rank(:display_order).all,
+             WebsiteFinder::Widget.new(widget).find.web_home_template].flatten
+    pages.map {|page| template_to_liquid(page)}
+  end
+  
+  def url_get_method
+    preview ? :preview_url : :url
+  end
+
+  def generated_url_1
```
variousred: @ceckert good find, these methods are not actually scoping to the location, and finding the first page whos slug matches.

ceckert: @variousred good. So I wasn't crazy. I just happened to test using the right location!

variousred:
@ceckert see https://github.com/G5/g5-content-management-system/commit/1f5fb6ca864bf8cb474b92d09402487da00e943a for the fix to the wrong urn being prefixed to the path

app/drops/widget_drop.rb
variousred: fixed

ceckert:
@variousred I tested and this is working on CR. :shipit: once :green_heart:

PR #592, opened 2015-03-17
Area pages 89930432

variousred:
https://www.pivotaltracker.com/story/show/89930432


variousred:
@levibrown @thescubageek @dreuss

public/area_page.js
```diff
@@ -1,10 +1,10 @@
 function app() {
```
dreuss: Any adversity to name-spacing this function to something like area_page_app() ?  app() seems really general, but maybe there's a reason for that naming, maybe this file is compiled from somewhere else & the name is auto-generated.  A name-spaced function would help any future developers with a little clue as to what/where the function does/exists.

variousred: @dreuss good point. And you're right, it does look like this was compile from coffee, but who knows where the source file is. @bbauer ? Either way, I'll change this to what you suggest Dave.

bbauer: Good question. Looking...

PR #595, opened 2015-03-19
Change url helper to not add a protocol

quady:
I ran into an error where it was trying to pull a form across over http
on an https page and causing errors. So now it will just use the page's
protocol and hopefully avoid that issue in the future.

### DIFF
--- a/lib/client_services.rb
+++ b/lib/client_services.rb
@@ -38,8 +38,7 @@ class ClientServices
     end
 
     define_method("#{service}_url") do |secure: false|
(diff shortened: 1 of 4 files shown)


variousred:
awesome, :shipit:

levibrown:
:+1:

jcstringer:
@quady I'm going to default to always use https for the service names in a constellation so this PR should no longer be needed. I'm gonna close it out just to keep the house clean.

PR #598, opened 2015-03-26
preview helper method is not actually only for preview. fixes that

variousred:
@ceckert this will fix some cta links rendering in preview format on CLWs


ceckert:
nice :icecream: :art:

PR #599, opened 2015-03-31
set show sub nav to true when child pages have `nil` which really just m...

variousred:
set show sub nav to true when child pages have `nil` which really just means not false

@ceckert this will fix the issue for sub navigations not showing without modifying all the settings


ceckert:
:shipit:

PR #605, opened 2015-04-09
Client services fix

thescubageek:
Updates Client Services to use G5HerokuAppNameFormatter instead of it's own BS way of hacking together URLs, which is currently causing certain edge case Client URNs to explode.


thescubageek:
@jcstringer @riverblufftrail @variousred I pulled my fix out my branch and put it here so we can go ahead and move along with this....

bbauer:
:+1: Nice work

PR #607, opened 2015-04-13
[WIP] Single location deploys

ceckert:
Updates Static Website Compiler to allow websites git history to show true changes (removes force push) and supports ability to deploy single location for Single Domain client without removing all other location information.

Need to Consider:
1) What is a location is disabled, how is that removed from the Single Location Website?
2) Ability to revert on a location by location basis, instead of the entire Single Domain site.
3) UI changes needed to support this.
4) Testing.


lib/static_website/deployer.rb
```diff
@@ -45,31 +46,49 @@ def deployer
...
+          github_repo: source_repo,
+          heroku_app_name: @website.heroku_app_name,
+          heroku_repo: @website.heroku_repo,
+          git_url: @website.github_repo,
+          name: @website.heroku_app_name,
+          heroku_organization_name: @website.client.organization,
       }
     end
 
+    def single_domain_client?
```
ceckert: @variousred I have this in here as well as in lib/static_website/compiler/website.rb. How can we DRY that out?

lib/static_website/compiler/website.rb
```diff
@@ -19,26 +19,22 @@ def initialize(website)
...
+        if (!single_domain_client? or website.owner.corporate?)
+          htaccess.compile
+          sitemap.compile
+          robots.compile
+        end
       end
 
-      def clean_up
-        compile_directory.clean_up
+      def single_domain_client?
```
ceckert: @variousred See my other comment

variousred:
@ceckert re: domain type of client —I think these can be moved upstream and passed in as an argument

ceckert:
Moved this to https://github.com/G5/g5-content-management-system/pull/618

PR #610, opened 2015-04-17
Alternate Theme-based Layouts

thescubageek:
Exposes a hook for Themes from the Theme Garden to define their own custom layout files, which takes precedent over the selected value from the Layout Garden for layout.

This will enable our Theme Developers to create custom headers, footer, and content wrappers for Custom Clients on a per-theme basis.


thescubageek:
Still need to fix a failing spec before this ready for prime time.

PR #611, opened 2015-04-20
WIP: Code climate fixes

thescubageek:


thescubageek:
@levibrown Specs pass now :+1:

app/controllers/api/v1/seeders/content_stripes_controller.rb
```diff
@@ -21,19 +21,28 @@ def seeder_params
 
   private
 
+  def is_content_stripe?(cs)
```
mmitchellg5: this logic already lived on the model, why bring it into the controller?

app/controllers/api/v1/seeders/content_stripes_controller.rb
```diff
@@ -21,19 +21,28 @@ def seeder_params
...
-    # - example:  params = { slug: 'apply', index: 2 }
     index = params[:index].try(:to_i) || 1
-    cs = web_template.widgets.select { |w| w.is_content_stripe? } if web_template
+    get_widget_by_index(index)
+  end
+
+  def get_widget_by_id(id)
+    w = Widget.find(id)
+    is_content_stripe?(w) ? w : nil
+  end
```
mmitchellg5: seems like these two get_widget methods are more like scopes. This one is named a bit misleadingly, the implied contract is that given an id it will return a widget, but this is rejecting anything thats a content stripe as well

app/controllers/api/v1/seeders/seeder_controller.rb
```diff
@@ -8,11 +8,12 @@ def index
...
 
   def show
+    pattern = params[:id].downcase.underscore
+    prefix = /^#{pattern}/
     response = serializer.new(nil).get_yaml_files.inject([]) do |arr, file|
-      prefix = /^#{params[:id].downcase.underscore}/
       arr << file unless prefix.match(file).blank?
       arr
-    end unless serializer.blank?
+    end unless serializer.blank? || pattern.blank?
```
mmitchellg5: pattern will aways exist or we wouldnt hit the show method here right?

app/models/concerns/widgets/content_stripe_widget.rb
```diff
@@ -25,10 +25,6 @@ def layout_count_values
       }
     end
 
-    def is_content_stripe?
-      true
-    end
```
mmitchellg5: why did you remove these?

app/models/setting.rb
```diff
@@ -46,6 +46,10 @@ def is_layout?
     value == 'Content Stripe' || value == 'Column'
   end
 
+  def value_or_default
+    value || default_value
```
mmitchellg5: isnt this what best value does already

app/controllers/api/v1/base_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def klass_name
+    controller_name.singularize
+  end
+
+  def klass_params
+    raise NotImplementedError
+  end
+end
```
mmitchellg5: good stuff

app/models/web_home_template.rb
```diff
@@ -12,7 +12,7 @@ def compile_path
   end
 
   def preview_url
-    "#{owner.urn}/#{vertical}/#{state}/#{city}"
+    "#{owner.urn}/#{client.vertical_slug}/#{owner.state_slug}/#{owner.city_slug}"
```
mmitchellg5: what is this change doing

app/models/web_template.rb
```diff
@@ -22,13 +18,9 @@ class WebTemplate < ActiveRecord::Base
     to: :website, allow_nil: true
 
   validates :title , presence: true
-  validates :name  , presence: true
+  validates :name  , presence: true , uniqueness: { scope: :website, case_sensitive: false }
```
mmitchellg5: nice

app/models/web_template.rb
```diff
@@ -22,13 +18,9 @@ class WebTemplate < ActiveRecord::Base
...
   validates :slug  , presence: true ,
-    format: {
-      with: /\A[-_A-Za-z0-9]*\z/,
-      message: "can only contain letters, numbers, dashes, and underscores."
-    },
-    unless: :new_record?
+    format: { with: /\A[-_A-Za-z0-9]*\z/, message: "can only contain letters, numbers, dashes, and underscores." }, unless: :new_record?
```
mmitchellg5: liked this split up because now i have side scrolling on github ;)

app/models/web_template.rb
```diff
@@ -39,11 +31,7 @@ class WebTemplate < ActiveRecord::Base
   scope :created_at_asc, -> { order("created_at ASC") }
   scope :top_level, -> { where(parent_id: nil)}
 
-  before_validation :default_enabled_to_true
-  before_validation :default_in_trash_to_false
-  before_validation :default_title_from_name
-  before_validation :default_slug_from_title
-
+  before_validation :set_defaults
```
mmitchellg5: sweet

app/models/web_template.rb
```diff
@@ -55,54 +43,22 @@ def meta_description_widgets
     widgets.meta_description
   end
 
-  def client
-    Client.first
-  end
```
mmitchellg5: I wonder if we could just set up a constant on boot for CLIENT because i'm sure this isnt the only place this can be removed from

app/models/web_home_template.rb
mmitchellg5: i get it now, these methods have been removed from the model, which does make sense to me

app/models/web_template.rb
```diff
@@ -112,37 +68,34 @@ def web_page_template?
...
 
   def javascripts
     show_javascripts + lib_javascripts + website_template_javascripts
   end
 
-  def website_compile_path
-    website.compile_path if website
+  def web_template_layout_html
+    web_theme_layout_html || website_layout.html
   end
```
mmitchellg5: the name of this is a little bit confusing to me, its still just the layout_html, not specific to a particular web_template

app/models/web_template.rb
```diff
@@ -112,37 +68,34 @@ def web_page_template?
...
-  def website_colors
-    website.colors if website
-  end
-
-  def website_fonts
-    website.fonts if website
+  def web_theme_layout_html
+    garden_theme = website_template.try(:web_theme).try(:garden_web_theme)
+    begin
+      open(garden_theme.layouts[0]).read unless garden_theme.layouts[0].blank?
```
mmitchellg5: was this model reading from remote before? this feels bad

app/models/web_template.rb
```diff
@@ -112,37 +68,34 @@ def web_page_template?
...
+    rescue OpenURI::HTTPError
+    end if garden_theme
   end
 
   def stylesheets_compiler
-    @stylesheets_compiler ||=
-      StaticWebsite::Compiler::Stylesheets.new(stylesheets,
-      "#{Rails.root}/public", website_colors, website_fonts, owner_name, true)
+    @stylesheets_compiler ||= StaticWebsite::Compiler::Stylesheets.new(stylesheets, "#{Rails.root}/public", website.try(:colors), website.try(:fonts), owner.try(:name), true)
```
mmitchellg5: ahhhhhh side scroling!

app/models/web_template.rb
```diff
@@ -112,37 +68,34 @@ def web_page_template?
...
   end
 
   def stylesheet_link_paths
-    LOGGERS.each{|logger| logger.debug("\n#### sending compile to stylesheets_compiler for web_template #{name}\n")}
+    write_to_loggers("\n#### sending compile to stylesheets_compiler for web_template #{name}\n")
```
mmitchellg5: very nice

app/models/web_template.rb
```diff
@@ -150,17 +103,15 @@ def preview_stylesheet_link_paths
   end
 
   def javascripts_compiler
-    @javascripts_compiler ||=
-      StaticWebsite::Compiler::Javascripts.new(javascripts,
-        "#{Rails.root}/public", name, owner_name)
+    @javascripts_compiler ||= StaticWebsite::Compiler::Javascripts.new(javascripts, "#{Rails.root}/public", name, owner.try(:name))
```
mmitchellg5: I dont have a touch mouse!! ahhhh

app/models/web_template.rb
```diff
@@ -197,8 +148,7 @@ def body_class
   end
 
   def location_body_class
-    corporate = website.owner.try(:corporate)
-    corporate ? "site-corporate" : "site-location"
+    website.corporate? ? "site-corporate" : "site-location"
```
mmitchellg5: heh nice, this was very strange before

app/models/website_template.rb
```diff
@@ -3,39 +3,43 @@ def all_widgets
     widgets.not_meta_description
   end
 
+  def drop_target_widgets(html_id)
+    drop_targets.where(html_id: "drop-target-#{html_id}").first.try(:widgets) || []
+  end
```
mmitchellg5: what if this was called 
```def widgets_for_drop_target_type(html_id)```

File: app/models/widget.rb [hunk 19]
```diff
@@ -147,28 +115,18 @@ def parent_widget
...
-
-  def set_child_widget(position, widget)
-    nil
-  end
-
   def liquid_render(html, params)
-    Liquid::Template.parse(html).render(params)
+    begin
+      Liquid::Template.parse(html).render(params)
+    rescue => e
```
mmitchellg5: do you know the error class for a liquid rendering error?

lib/saves_manager.rb
```diff
@@ -40,6 +36,28 @@ def restore(save_id)
...
+      { id: parse_bucket_name(leaf).rpartition('.').first,
+        created_at: leaf.object.last_modified }
+    end.sort {|a,b| b[:created_at] <=> a[:created_at]}
+  end
+
+  def sort_bucket_children(limit=SAVES_LIMIT)
+    bucket_children.map do |leaf|
+      { leaf: leaf, file_name: strip_bucket_name(leaf)}
+    end.sort { |a,b| b[:file_name] <=> a[:file_name] }.take(limit)
+  end
```
mmitchellg5: you can do the name, date sort all in one method
```some_array.sort {|a,b| [b[:created_at], b[:file_name]] <=> [a[:created_at], a[:file_name]]}```

File: app/controllers/api/v1/seeders/content_stripes_controller.rb (hunk 1 above)
Comment (thescubageek) [inline app/controllers/api/v1/seeders/content_stripes_controller.rb]: Legacy crap, might as well pull it out

**Comment by mmitchellg5** [comment]
re: the compiler/deployer changes, ensure these are all tested extensively in the wild as this code is quite perilous

File: lib/saves_manager.rb [hunk 21]
```diff
@@ -1,5 +1,6 @@
 class SavesManager
   FOLDER_NAME = "db-backups"
+  SAVES_LIMIT=100
```
mmitchellg5: this limit implementation is good, however, it will a larger issue of storing too many saves, costing storage.

app/controllers/api/v1/seeders/content_stripes_controller.rb
thescubageek: It is only returning anything that is a content stripe (not rejecting), but yeah, it should be named get_content_stripe_by_id (etc)

app/controllers/api/v1/seeders/seeder_controller.rb
thescubageek: true, redundant

app/models/setting.rb
thescubageek: No best value interjects most recently entered value between the current and default value. So if we want everything to default to a certain liquid, but someone enters something else on another (same type) widget on the site, it will no longer default to the liquid but to the last value they entered. Add this gives us a simpler way to doing either value or default while cutting out all the wonky best value logic.

app/models/web_template.rb
thescubageek: Where would this be defined?

app/models/web_template.rb
thescubageek: Yes it was, it justs pull it in from the layout from the external Layout Garden instead. Once our Gardens are living in the CMS this will be a non-issue. Sucks both ways until we elimiate that endpoint

app/models/website_template.rb
thescubageek: I like that better, renaming it.

app/models/widget.rb
thescubageek: No but I should be more explicit with the error class

lib/saves_manager.rb
thescubageek: Yeah but I don't want to because created_at will actually hit the AWS bucket where as file_name will attempt to glean the date from the name alone (no bucket hit required). We can then cull the list (the .take(limit)) and then sort them again by date modifed as a sanity check (haven't confirmed if this is necessary). This way we don't hit AWS more times than the SAVES_LIMIT, which was what has been killing our clients with tons of saves.

lib/saves_manager.rb
thescubageek: Agreed, it only masks the issue, doesn't solve it, but at least "bad" clients can still load their saves page. Culling the buckets is a bigger task.

app/models/setting.rb
mmitchellg5: it seems like this should replace best value, or best value should be refactored to do the right thing, otherwise there are two methods that appear to be doing very similar things, and both will need to be maintained, devs will need to know which to use where and why

thescubageek:
This branch is deprecated, but I'm keeping it around for reference on future refactors

PR #615, opened 2015-04-24
fix problem with single domain redrirects

ceckert:
@variousred Discovered this yesterday with Virginia Self Storage. I have tested with their CMS and it is working as intended.


mmitchellg5:
:shipit:

PR #623, opened 2015-05-13
add light and dark options for tertiary color

lance-hardy:
Just adding light and dark variables in boilerplate for the tertiary color.


crissmancd:
:shipit:

PR #629, opened 2015-05-22
dont create directory if it exists

ceckert:
Defect with corp location deploys. @thescubageek PR please?


thescubageek:
:+1: once we get specs to "pass" (requires hitting refresh on integration specs)

thescubageek:
Looks good to :ship:

PR #631, opened 2015-05-26
fix bug in client liquid parameters for go squared

thescubageek:
Fixes bug that was causing client_go_squared_id liquid to be nil


ceckert:
:shipit: when :green_heart:

PR #633, opened 2015-05-27
Single domain url fix

ceckert:
Need to use Client domain for single domain clients. Zip code search uses the Location url.


thescubageek:
:+1: :ship:

PR #634, opened 2015-05-27
Fix Header Notifications

bbauer:
- [x] Set notifications url
- [x] Update g5_header for latest fixes on it's end


bbauer:
@ceckert I'd like to get this tested on a staging CMS before releasing. Is there a test CMS setup somewhere I could use?

ceckert:
I'm testing locally. Some minor stylistic differences from DSH. The svg next to icon is not displaying, and the exclamation mark and box border in Notifications is slightly different. 

image

image

bbauer:
@jonnysamps?

jonnysamps:
@bbauer @ceckert I think you need to hard refresh your browser. I've seen that happen a lot with the dashboard.

ceckert:
@jonnysamps no. Tried that and a different browser. Still same issue.

PR #635, opened 2015-05-27
Bring in G5 Updatable

jcstringer:
https://www.pivotaltracker.com/story/show/94445022

This one's gonna need a lot of QA and peer review. Master has been merged in as of 5/27/15.


app/controllers/api/v1/locations_controller.rb
```diff
@@ -1,8 +1,8 @@
 class Api::V1::LocationsController < Api::V1::ApplicationController
   def index
-    render json: Location.all
+    render json: G5Updatable::Location.all, each_serializer: LocationSerializer
```
thescubageek: nice, never seen each_serializer

app/decorators/location_decorator.rb
```diff
@@ -10,6 +10,10 @@ class LocationDecorator < Draper::Decorator
     :service_1, :service_2, :service_3, :service_4, :service_5,
     :website
 
+  def self.object_class
```
thescubageek: What does this do?

app/models/clients/multi_domain_client.rb
```diff
@@ -1,5 +0,0 @@
-class MultiDomainClient < Client
```
thescubageek: Nice!

app/models/concerns/after_update_set_path_settings.rb
```diff
@@ -8,7 +9,8 @@ module AfterUpdateSetPathSettings
   private
 
   def should_update_location_paths?
-    city_changed? || state_changed?
+    observed_properties = %w{city state}
```
thescubageek: Much better

app/models/concerns/after_update_set_setting_locations_navigation.rb
```diff
@@ -9,11 +10,12 @@ module AfterUpdateSetSettingLocationsNavigation
   private
 
   def should_update_locations_navigation_settings?
-    state_changed? || city_changed? || name_changed?
+    observed_properties = %w{city state name}
+    properties_changed? && property_changed?(observed_properties)
```
thescubageek: Any way we can DRY up the three repetitions of the follow?:
```
properties_changed? && property_changed?(observed_properties)
```

app/decorators/location_decorator.rb
jcstringer: The decorator assumes it should use a class of the same name, in this case we want it to actually use G5Updatable::Location. You can over-ride the method per decorator to do this.

app/models/concerns/location_liquid_parameters.rb
```diff
@@ -3,7 +3,7 @@ module LocationLiquidParameters
 
   def location_site_liquid(l_location)
     site_info = {
-      "website_urn"               => l_location.website.urn,
+      "website_urn"               => l_location.reload.website.urn,
```
thescubageek: Why is this reload necessary?

lib/color_finder.rb
```diff
@@ -35,6 +35,6 @@ def web_theme
   end
 
   def valid?
-    @location && @location.website && @location.website.website_template
+    @location.try(:reload) && @location.reload.website && @location.website.website_template
```
thescubageek: We need these reloads because...?

jcstringer:
@thescubageek the .htaccess specs are commented out at the advice of @levibrown who has already redone these specs in another PR that will be merged in soonish. So rather then fix them I'm moving them out of the way, they were writing to the file system multiple times on every run.

PR #636, opened 2015-05-28
Update liquid

thescubageek:
* Updates NAE and VLS liquid variables to use centralized endpoints
* Adds g5_ops gem, which adds a version file
* Updates docs in Ember app


johnlucia:
:+1:

PR #643, opened 2015-06-02
G5 updatable mergefest

thescubageek:
Updatable Mergefest continues...


jcstringer:
@ceckert @mmitchellg5 @thescubageek @levibrown  guys I'm getting sick of these integration specs that seem to fail 75% of the time. Long term we need to fix them or better yet move the tests into the ember app that is the front end for this app anyways, so I'm wondering if you guys are ok with my commenting out the problematic ones. I'd like circleCI to be dependable when it tells us the state of the test suite on a topic branch for this app.

PR #646, opened 2015-06-03
Spec tacular

jcstringer:
@thescubageek here are the specs in all their glory. Please review, I made some assumptions about removing some try statements and assuming things like there would always be a theme, etc. Also did some minor refactoring, couldn't resist a bit while I was in there.


app/models/concerns/client_liquid_parameters.rb
```diff
@@ -1,14 +1,18 @@
 module ClientLiquidParameters
   extend ActiveSupport::Concern
 
+  included do
```
thescubageek: Nice, much cleaner.

app/models/concerns/client_liquid_parameters.rb
```diff
@@ -17,22 +21,10 @@ def client_dxm_liquid(l_client)
...
     }
-    p = %w(cls_url cxm_url dsh_url cpas_url)
-    dxm_info.merge(load_client_liquid(l_client, p))
+    dxm_info.merge(load_liquid(l_client, CLIENT_DXM_LIQUID, "client"))
   end
 
   def client_service_liquid(l_client)
-    p = %w(service_1 service_2 service_3 service_4 service_5)
-    load_client_liquid(l_client, p)
+    load_liquid(l_client, LiquidParameters::SERVICE_LIQUID, "client")
```
thescubageek: Sweet

app/models/concerns/liquid_parameters.rb
```diff
@@ -63,15 +66,19 @@ def client_liquid(l_client=nil)
...
+    liquid_web_theme = liquid_web_template.web_theme
+    theme_liquid = {
+      "theme_name"                => liquid_web_theme.name,
+      "theme_slug"                => liquid_web_theme.garden_web_theme.slug,
+    }.merge(load_liquid(liquid_web_theme, THEME_LIQUID, "theme"))
+  end
+
+  protected
+
+  def load_liquid(object, params_list, prefix)
```
thescubageek: Always meant to get around to DRYing this one up, thanks.

app/models/concerns/location_liquid_parameters.rb
```diff
@@ -1,50 +1,55 @@
...
 
   def location_service_liquid(l_location)
-    p = %w(service_1 service_2 service_3 service_4 service_5)
-    load_location_liquid(l_location, p)
+    load_liquid(l_location, LiquidParameters::SERVICE_LIQUID, "location")
   end
 
   protected
 
   def load_location_liquid(l_location, params_list)
```
thescubageek: We can get rid of this method now since you've refactored it into load_liquid()

app/models/concerns/liquid_parameters.rb
```diff
@@ -63,15 +66,19 @@ def client_liquid(l_client=nil)
...
+    params_list.inject({}) do |h, p|
+      h["#{prefix}_#{p}"] = object.send(p)
```
thescubageek: I had object.try(p) here because I wanted the value to be set to nil if not found, rather than throwing an error. But maybe we do want the error behavior so we can detect broken liquid variables easier....

app/models/concerns/location_liquid_parameters.rb
```diff
@@ -1,50 +1,55 @@
 module LocationLiquidParameters
   extend ActiveSupport::Concern
-
+  included do
+    LOCATION_INFO_LIQUID     = %w(name name_slug phone_number office_hours access_hours floor_plans primary_amenity qualifier primary_landmark)
+    LOCATION_ADDRESS_LIQUID  = %w(city city_slug state state_slug state_name state_name_slug street_address neighborhood neighborhood_slug postal_code)
+    LOCATION_SITE_LIQUID     = %w(uid urn url relative_url domain)
+    LOCATION_SOCIAL_LIQUID   = %w(ga_tracking_id ga_profile_id)
+    LOCATION_SOCIAL_MAPPINGS = {
```
thescubageek: :+1:

spec/models/concerns/liquid_parameters_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        its([:page_slug]){ should eq web_template.slug } 
+        its([:page_created]){ should eq web_template.created_at.to_s.gsub(' UTC','') }
+        its([:page_updated]){ should eq web_template.created_at.to_s.gsub(' UTC','') }
+      end  
+
+      context "location_liquid" do
+        context "location_site_liquid" do
+          its([:website_urn]){ should eq website.urn }
+          its([:location_corporate]){ should eq location.corporate?.to_s }
+          described_class::LOCATION_SITE_LIQUID.each do |attr|
```
thescubageek: Good use of that class-level constant to keep the specs _much_ cleaner

thescubageek:
Looks really good @jcstringer , I really like the way you set up most of the liquid attributes on class-level constants and then used those constants to make very tight specs. Let's get rid of the deprecated load_location_liquid() method and then we can :ship:

jcstringer:
Gonna close this and re-open against master.

PR #647, opened 2015-06-08
Spec tacular

jcstringer:
Adding specs for the liquid logic that ended up causing us some pain with the mergefest party.


thescubageek:
Looks great @jcstringer, good refactoring of the liquid stuff. Thanks for putting the specs together

thescubageek:
:+1: :ship:

thescubageek:
@jcstringer Let's remember to bump the subversion to 1.1.3 before merging.

thescubageek:
@jcstringer Merge and deploy once you have approval from QA.

app/models/concerns/liquid_parameters.rb
```diff
@@ -3,10 +3,9 @@ module LiquidParameters
   include LocationLiquidParameters
   include ClientLiquidParameters
 
-  def extend_liquid_methods
-    liquid_parameters.each do |key, value|
-      define_singleton_method(key) { liquid_parameters[key] }
-    end
+  included do
+    SERVICE_LIQUID = %w(service_1 service_2 service_3 service_4 service_5)
```
variousred: these are methods defined on the including object?

jcstringer: These are just strings that are part of a constant on the including class, they get used later in some logic that dynamically builds some liquid methods onto locations and clients.

PR #649, opened 2015-06-08
make g5_updatable concerns use resque jobs

thescubageek:
Adds Resque jobs for the three concerns on g5_updatable_locations


ceckert:
:+1:

variousred:
made some comments on the merge, Resque.enqueue is a class method not an instance method so the stubs created in the specs for enqueue wont actually do anything

thescubageek:
@variousred Thanks I'll patch them up.

PR #652, opened 2015-06-09
Canonical updatables

jcstringer:
DRYs up a bunch of different places in the CMS that point to the different services in a constellation. Use the g5_heroku_app_name_formatter as the canonical place to ask a constellation for its services names, urns and urls.


lib/client_services.rb
```diff
@@ -1,8 +1,15 @@
 class ClientServices
-  SERVICES = %w(c cms cpas cls cxm dsh)
+  SERVICES = G5HerokuAppNameFormatter::UpdatableUrlFinder::UPDATABLE_APPS.push(:c)
```
variousred: whats the C service?

variousred:
nice :eyeglasses:  



(specs)

jcstringer:
@variousred the client_services class needed to know about the urn for just the client itself, i.e. "g5-c-i76q1xel-qa-test-beds-m". This comes into play for the client_app_name method. In reality this is just the same thing as the "client_urn" attribute on the G5Updatable::Client and client_app_name isn't even used anyplace that I can find, but I think I removed it and noticed something broke so I just left support for it intact.

thescubageek:
Looks good :ship: on Monday

thescubageek:
@jcstringer Merging this into master and kicking it out the door...

lib/client_services.rb
thescubageek: Client... it's how the URN is referenced in the HUB link, eg https://g5-hub.herokuapp.com/clients/g5-c-i8q4q20e-a1-locker-rental-and-self-storage-client

PR #653, opened 2015-06-09
Fix sd app name

thescubageek:
Fixes a bug where Client.first.website is nil in current master. This *does not* address the reason that website is nil (need to pair with @jcstringer on that), but at least provides a workaround to the 500 error that is currently preventing brand new SD Clients from working.


jcstringer:
Seems like we should not need this change if the issue was just bad data?

PR #655, opened 2015-06-11
Bootstrap updates

jcstringer:
* Don't define constants if they are already defined
* Attempt to squash a bug where the initializer for G5Updatable doesn't work as expected

https://www.pivotaltracker.com/story/show/96831420


thescubageek:
Bump subversion and :ship:

PR #659, opened 2015-06-19
if there is a corporate navigation widget prefer that over corporate …

variousred:
if there is a corporate navigation widget prefer that over corporate twebsite navigation setting


bbauer:
:+1: when :green_heart:

PR #672, opened 2015-07-02
Layout widget patch

thescubageek:
Migration patches bug in content stripes and column name settings.


thescubageek:
I hate merging my own PR but I sat here with @dominicalbano and we personally QAed it.... :-\

PR #673, opened 2015-07-02
patch nil bug on malformed widgets

thescubageek:
Patches nil bug on previous hotpatch


thescubageek:
Another hotpatch, another round of guilty commits with Garrett and Dom's approval

PR #679, opened 2015-07-08
Bumped up g5_ops version and added initializer

jonnysamps:


thescubageek:
@jonnysamps Get those specs to pass and we should be fine

PR #682, opened 2015-07-08
Active Anchor Color Only Applied to Top Level Navigation

aborows:
https://www.pivotaltracker.com/n/projects/1282248/stories/97919838

Before (active navigation item has dark color on dark background-color):
screen shot 2015-07-08 at 1 31 44 pm

After (active navigation item has light color on dark background-color):
screen shot 2015-07-08 at 1 33 04 pm


thescubageek:
This has been added into https://github.com/g5search/g5-content-management-system/pull/681

PR #692, opened 2015-07-09
2.0.3: Patch Content Stripe and Column to work with local Ember app

thescubageek:
add window.EMBER_HOST to layout widgets so it works with local Ember app


thescubageek:
This has been merged into 2.0.3

PR #693, opened 2015-07-09
2.0.3 alert fix

thescubageek:
Fixes bug where editing inside a Column or Content Stripe widget causes the notification alerts in the header to vanish


thescubageek:
This has been merged into https://github.com/g5search/g5-content-management-system/pull/688

PR #698, opened 2015-07-13
set default value of fade delay

crissmancd:


thescubageek:
:+1: This has been merged into https://github.com/g5search/g5-content-management-system/pull/688

PR #704, opened 2015-07-14
 Ppc

johnlucia:
This paves the way for the dynamic display of PPC numbers, and does a little performance tuning. Here are all the things this change does: 

- Currently we make an ajax call for phone numbers on every page load. We shouldn't need to do that, and it's hard on the centralized phone service. Use session storage to remember the json response from the first request, so we can skip that call on everything after the user's first page view.

- If a visitor lands on the site via a PPC ad, then a query param of 'cpm' will be added to the URL. (Example: www.myghettoapartments.com?cpm=bingppc) In this case we should display the locations PPC number associated with the 'bingppc' code.

- If the user clicks around the site, the 'cpm' query param will go away. But, we still need to keep showing the ppc number.

- The old stuff assumes a location will only ever have 1 PPC number, but that's not the case. If a user leaves the site, and returns via a different PPC campaign, we need to display that new PPC number instead.


public/javascripts/libs/phone-service.coffee
```diff
@@ -117,4 +78,92 @@ class PhoneNumber
...
+    @ppcNumber = @getOrStorePpcNumber()
+    @mobileNumber = @loc['mobile_number']
+    @defaultNumber = @loc['default_number']
+
+
+  dynamicNumber: ->
+    return @ppcNumber if @ppcNumber
+    return @mobileNumber if @mobileNumber && $(document).width() < @MOBILE_WIDTH
+    return @defaultNumber if @defaultNumber
+    return ""
```
dreuss: Well thought out about the chain of events / cases of how the number is derived.

public/javascripts/libs/phone-service.coffee
```diff
@@ -1,8 +1,3 @@
-# make these global so we don't waste AJAX calls on multiple widgets
-window.phoneNumberData = null
-window.phoneNumberAjax = false
-window.phoneNumberError = false
-
```
dreuss: Able to ditch these since you store the cpns request in a session variable that gets used on subsequent pages, instead of having to hit the service on every page load. Nice.

dreuss:
Overall looks like a well thought out, robust refactor, which should be more performant for our CPNS reducing ajax calls for every page load.  As this will affect pre-configured existing widgets, I'd say QA will have their work cut out for them testing the places where this phone number service is implemented.  As long as QA gives you the thumbs up, and the swapping of numbers has the consistent and expected results, I'd say go for it.  Any other engineers who'd like to add thoughts here please do. :+1:

thescubageek:
Great work @johnlucia :+1: :ship: This has been merged into 2.0.7-merge: https://github.com/g5search/g5-content-management-system/pull/731

PR #714, opened 2015-07-16
2.0.5 edit loc map

thescubageek:
* Adds edit.js for Corporate Locations Map widget, which hides/shows fields depending on the target widget type setting

https://www.pivotaltracker.com/story/show/99004694


public/static/components/widgets/corporate-locations-map/javascripts/edit/edit.js.coffee
```diff
@@ -0,0 +1,33 @@
...
+      $('.corp-search-page').attr('placeholder', 'locations/') if changed
+    else if @target == 'corp-search'
+      cityStateZipFields.show()
+      corpSearchFields.show()
+      $('.corp-search-page').attr('placeholder', 'apartment-search/') if changed
+    return
+
+  hideAndClear: (fields)->
+    fields.hide()
+    fields.find('input[type=text]').val("")
```
dreuss: Nice thought here, good call to clear out values so things aren't getting submitted that we don't want.

dreuss:
Pretty nice way to ease the workflow, and show fields conditionally based upon how a user wants to configure the widget.  Nice work, I'm sure this pattern will get implemented on some more widgets.  :+1:

thescubageek:
This has been merged into 2.0.5-merge: https://github.com/g5search/g5-content-management-system/pull/712

PR #723, opened 2015-07-21
2.0.6 spinup callbacks

thescubageek:
Disable nav settings callbacks when running simple spinup


thescubageek:
@mmitchellg5 I _think_ this will fix our UpdateNavSettingsJob issue with spinup..... I'll get this tested tomorrow.... should make Spinup Saves a lot less stressful on Resque

applist.txt
```diff
@@ -1,131 +1,138 @@
-g5-cms-1sbdk8wp-bb-self-storag
```
thescubageek: Ignore this file, it's just used for master deploys and was updated by the most recent push... has no effect on the actual CMS

variousred:
nice :shipit:

thescubageek:
@mmitchellg5 Made a few more tweaks after running the test against Byron's CMS.... found that the update nav settings job was also bound to the Web Template model.... cleaned that out.

Also, added a new job that updates both themes and widgets..... this will be used by the post deploy task (instead of two jobs, one for themes and one for widgets) since it will consume fewer resources

Tested this both locally and against Byron's test CMS, confirmed it works.

variousred:
:shipit:

variousred:
fix tests

thescubageek:
Fixed the spec, got the :green_heart:

thescubageek:
@variousred This has been merged into 2.0.6-merge: https://github.com/g5search/g5-content-management-system/pull/725

PR #726, opened 2015-07-22
2.0.X liquid extensions

thescubageek:
Extends liquid variables as requested in this PR: https://www.pivotaltracker.com/story/show/99304864

The following liquid variables have been added:

location_neighborhood_2
location_google_plus_id
location_nearby_schools
location_nearby_employers
location_floor_plans
location_apartment_amenity_1
location_apartment_amenity_2
location_community_amenity_1
location_community_amenity_2
location_landmark_1_type
location_landmark_1_name
location_landmark_2_type
location_landmark_2_name
location_property_feature_1
location_property_feature_2
location_property_feature_3
location_foursquare_id
location_tumblr_id
location_vimeo_id


jcstringer:
:ship:

jcstringer:
Just remember the corresponding hub code will need to be re-deployed to production for this to ship (and I see it's WIP so you prob already thought of that). Otherwise these new methods will prob blow up because the old hub master is currently deployed to production. Come to think of it - we should validate that CMS'es which haven't been synced with G5Updatable and thus don't know about the new fields, won't blow up when you try to access them. If it turns out that is an issue, we can always create a rake task in the hub to kick of Updatable Syncs for all clients.

thescubageek:
This has been merged into 2.0.9-merge: https://github.com/g5search/g5-content-management-system/pull/748

PR #734, opened 2015-07-27
2.0.7 cf lane sg

sheena-gygax:
Pending Custom Theme
- Adds padding to stack images in a column
- Rms js function, use styles instead 
- Target first row col image to add correct class


thescubageek:
@sheena-gygax Your changes have been merged into 2.0.7-merge: https://github.com/g5search/g5-content-management-system/pull/731

PR #736, opened 2015-07-28
2.0.X config endpoints

thescubageek:
Sets up g5_ops config endpoint to report on Web Themes and Widgets for the CMS.

The config endpoint provides the following info about a CMS:
* version
* which locations are using what themes
* which locations and pages are using what widgets

The JSON from this endpoint is used by the G5 App Wrangler to data mine information about what's being used across all of our CMSs.

Since calculating the data for this endpoint is expensive, it is set up to be scraped via the ConfigUpdaterJob, which is triggered via the /config_update POST endpoint. The Wrangler will be schedule to POST to this endpoint nightly so that the CMSs update their config reports every 24 hours.

This is related to this PR for the G5 App Wrangler: https://github.com/g5search/g5-app-wrangler/pull/5


app/models/web_template.rb
```diff
@@ -37,6 +37,17 @@ class WebTemplate < ActiveRecord::Base
...
 
+  def all_nested_widgets
+    return @all_nested_widgets if @all_nested_widgets
+    top_widgets = widgets
+    nested_widgets = top_widgets.map { |w| w.widgets }.flatten
+    @all_nested_widgets = (top_widgets + nested_widgets).flatten
+  end
+
+  def all_garden_widgets
+    all_nested_widgets.map { |w| w.garden_widget }.uniq
```
jcstringer:
```
all_nested_widgets.map(&:garden_widget).uniq
```

app/models/widget_config.rb
```diff
@@ -0,0 +1,39 @@
+class WidgetConfig < ActiveRecord::Base
+  validates :location , presence: true
+  validates :location_slug, presence: true
+  validates :urn  , presence: true , uniqueness: true
+  validates :widget_count, presence: true
+  validates :garden_widgets, presence: true
+  validates :garden_widget_slugs, presence: true
+
```
jcstringer: DRY up validations that are common to multiple classes into a Concern and you'll get a better code climate score.

app/models/widget_config.rb
```diff
@@ -0,0 +1,39 @@
...
+    self.update_widget_page_configs(config, website)
+    config
+  end
+
+  def self.update_widget_page_configs(config, website)
+    config.widget_page_configs.destroy_all
+    page_configs = website.web_pages.map do |p|
+      WidgetPageConfig.update_config(config, p)
+    end
+    page_configs
```
jcstringer: You don't actually need to explicitly return this, just have the call to map be the last thing in this method and it will be returned.

jcstringer:
This is looking good, just need some specs

thescubageek:
Merged into 2.0.8: https://github.com/g5search/g5-content-management-system/pull/740

PR #740, opened 2015-07-28
WIP: 2.0.8 Merge

thescubageek:
### CMS Architecture:
* Extends G5 Ops config endpoint to report on CMS's Web Themes and Widgets by Location and Page
* Adds WebThemConfig, WidgetConfig, and WidgetPageConfig models for recording CMS config data
* Opens update_config endpoint for triggering nightly config rake tasks via the G5 App Wrangler
* Patches minor bugs in WidgetEditorFilter for photo and gallery edit fields

### Widgets:
* HTML: Adds styles for ordered list `<ol>` tags
* Converts the following widgets to use the new WidgetEditorFilter: Analytics, City State Zip Search, Call Outs, Corporate Locations Map, Coupon, CSS Override, Directions, Directory, Featured Properties, and Footer Info

### Themes:
* Boilerplate 2: Adds Neat SCSS, refactors a number of partials and mixins, and makes Boilerplate more easily customizable for Theme Developers
* CF Lane: Patches Photo overlay to work with multiple photos, and adds a few minor CSS padding fixes
* Cobalt: Adds active state to nav menu, removes center align on headings
* Luxe: Fixes Windows Firefox bug that adds default arrows to select element
* Fusion: Adds margin-bottom CSS to Contact Info widget
* Welborne: Adds CSS styles for nested `<ul>` tags

PR #741, opened 2015-07-28
2.0.8 widget rampage

thescubageek:
Converts the following widget's edit.html to the new WidgetEditorFilter liquid:

* Analytics
* City State Zip Search
* Call Outs
* Corporate Locations Map
* Coupon
* CSS Override
* Directions
* Directory
* Featured Properties
* Footer Info


thescubageek:
This has been merged into 2.0.8-merge: https://github.com/g5search/g5-content-management-system/pull/740

dreuss:
The refactor on all the edit pages looks consistent, and great reduction in repetitive code.  I might have @mmitchellg5 look over some of the ruby codes just to have a second set of eyes, but looks good. :+1:

app/filters/widget_editor_filter.rb
```diff
@@ -33,7 +33,7 @@ def edit_integer_field(setting, label="", placeholder="", method=:value, klass="
     ).html_safe
   end
 
-  def edit_textarea_field(setting, label="", placeholder="", method=:value, klass="")
+  def edit_textarea_field(setting, label="", placeholder="", method=:value, klass="", cols=40, rows=3)
```
variousred: with this many settings, i might accept an options hash

thescubageek: Hashes are really cumbersome to pass via liquid.... you have to do some ugly formatting and then use JSON.parse on the other end to make it work (see the edit_select_field etc)

variousred: ah, gotcha

PR #749, opened 2015-07-30
2.0.9 html list styles

crissmancd:
HTML widget now has default list styles. Fix themes that are using images for list items.


thescubageek:
:+1: thanks for patching this up so quick. merged into 2.0.9: https://github.com/g5search/g5-content-management-system/pull/748

PR #751, opened 2015-07-30
adding default nav and active color

lance-hardy:
- nav color is always white
- nav should always have an active state of default color darkened


thescubageek:
@lance-hardy This has been merged into https://github.com/g5search/g5-content-management-system/pull/748

PR #752, opened 2015-07-30
2.0.9 cc

thescubageek:
Temp fix for Case Corp to get them their beloved Corp Contact form


quady:
@thescubageek I had to swap the form name to use an underscore because the dash did not play nice.  Could you also make that change on your liquid branch or would you like me to make a pull req into that branch?

thescubageek:
@quady Good to know, I was wondering why my current liquid conversion of this was failing. I'll make the patch

thescubageek:
@quady These changes went out with 2.0.9

PR #756, opened 2015-08-03
Update Analytics and Meta widgets, moving Optimizely to Analytics

ceckert:
I tested this locally and it seems to be working OK. @thescubageek can I get a review please?


public/static/components/widgets/analytics/index.json
```diff
@@ -17,6 +17,11 @@
   "settings": {
     "instance": [
       {
+        "name": "optimizely_id",
+        "editable": true,
+        "default_value": ""
```
thescubageek: Is Optimizely going to be a HUB value? If so, we should (eventually) set up location_optimizely_id as a liquid var and use value or default

public/static/components/widgets/analytics/show.html
```diff
@@ -1,3 +1,6 @@
+{% if widget.optimizely_id.value and widget.optimizely_id.value != blank %}
+<script type="text/javascript" src="//cdn.optimizely.com/js/{{ widget.optimizely_id.value }}.js"></script>
```
thescubageek: If we're using best_value above in the edit.html, we should use it here and in the code below as well.... that way the user doesn't even have to open the edit panel for the "best value" to set the value, instead it should work out of the box...... of course I'm in much bigger favor of using value_or_default rather than the voodoo of best_value, but that won't work for us until we have a default liquid value to use.

public/static/components/widgets/meta/edit.html
```diff
@@ -27,18 +27,6 @@
   </div>
 
   <div class="form-field">
-    {{ widget.optimizely_id.id_hidden_field }}
```
thescubageek: I'm going to be converting these changes into the new Edit liquid today, so I'll make sure to drop this field and merge in your changes.

public/static/components/widgets/analytics/index.json
ceckert: My understanding is no. This is not an "every site" type setting. Only used by CRO team periodically. 

Tyler, Scott, your thoughts?

ceckert: For now, only select clients are getting active CRO. Optimizely ID is based
upon the domain, not the location. So I'm not sure if that is "hubbable" or
not. Eventually, when we get *Google Tag Manager *configured, we will stop
asking ENGR for these types of requests as we can manage *all *the various
scripts for the client within GTM.



Tyler Herrick
TECHNICAL SEO MANAGER

<http://www.getg5.com> <http://www.getg5.com> <http://www.getg5.com>[image:
G5 Website] <http://www.getg5.com>
www.GetG5.com <http://www.getg5.com/>
T 541.647.1262 | F 541.225.5166

FOLLOW US <https://plus.google.com/u/0/101198449642176712699/about>
<https://www.linkedin.com/company/getg5> <https://twitter.com/GetG5>
<https://www.facebook.com/GetG5>

On Mon, Aug 3, 2015 at 9:59 AM, Christopher Eckert <
christopher.k.eckert@gmail.com> wrote:

> My understanding is no. This is not an "every site" type setting. Only
> used by CRO team periodically.
>
> Tyler, Scott, your thoughts?
>
> Sent from my iPhone
>
> On Aug 3, 2015, at 12:53 PM, Steve Craig <notifications@github.com> wrote:
>
> In public/static/components/widgets/analytics/index.json
> <https://github.com/g5search/g5-content-management-system/pull/756#discussion_r36105125>
> :
>
> > @@ -17,6 +17,11 @@
> >    "settings": {
> >      "instance": [
> >        {
> > +        "name": "optimizely_id",
> > +        "editable": true,
> > +        "default_value": ""
>
> Is Optimizely going to be a HUB value? If so, we should (eventually) set
> up location_optimizely_id as a liquid var and use value or default
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/g5search/g5-content-management-system/pull/756/files#r36105125>
> .
>
>

-- 
*This email may contain information that is privileged, confidential, or 
proprietary and is intended solely for the named addressee. If you are not 
the addressee, or if it appears that you have received this email in error, 
please advise me immediately by reply email, do not disclose, copy, or 
distribute the contents, and immediately delete the message and any 
attachments from your system. Thank you.*

thescubageek:
@ispyty @ceckert Makes sense to me, I'll get these changes merged into 2.0.9

thescubageek:
@ceckert This has been merged into 2.0.9-merge: https://github.com/g5search/g5-content-management-system/pull/748

PR #757, opened 2015-08-03
fix Windows FF bug

sheena-gygax:
Multifamily Search widget  bug for Windows Firefox 39.0

- this hides the browser default arrow button on the select element (to allow the styled arrows to show instead of 2 arrows)


thescubageek:
@sheena-gygax Merged into 2.0.9: https://github.com/g5search/g5-content-management-system/pull/748

PR #760, opened 2015-08-04
2.1.0 Merge

thescubageek:
### CMS Architecture:
* **New!** Serialized Saves: Location Saves now use the Deep Seeder Simple Spinup files instead of database backup files to save and restore websites - no longer will save files break after a database migration has been run (there is still one known bug where restored Navigation widgets will not have any pages select - this will be patched in an upcoming release)
* **New!** Theme Viewer: now uses a Grid view for showing all the Themes - much easier to select a Theme compared to the old scrolling system

### Widgets:
* Lead Form: Changes YVSL and Yardi Voyager Forms to be Intergrated Forms (the CLS handles serving the correct form) - fixes bug where certain YVSL forms were unselectable
* Converts the following widgets to use the new WidgetEditorFilter: Lead Form, Map, Meta, Meta Description, PDF Embed

### Themes:
* CF Lane: Minor CSS tweaks on `<h6>` tags and Call To Action widget
* Holland Corp: Adds new CSS styles for social icons, and significantly refactors JS and CSS
* Holland Corp: Adds color-changing header to home page only, plus minor CSS typography tweaks
* Luxe: Adds CSS style improvements for the for the Locations Navigation widget in Hero Content Stripe
* Welborne: Makes minor responsive CSS margin tweak for Navigation widget list items

PR #767, opened 2015-08-05
decrease margin-bottom on li if > min-large

sheena-gygax:
reduce margin-bottom a tad if >= min-large


thescubageek:
Merged, will :ship: with 2.1.0: https://github.com/g5search/g5-content-management-system/pull/760

PR #770, opened 2015-08-06
Saves navigation

variousred:
@thescubageek small refactoring on websiteseeder, before i get into the navigation seeding issue


variousred:
@levibrown

levibrown:
:+1:

lib/seeder/website_seeder.rb
```diff
@@ -18,24 +18,26 @@ def seed
...
     
-    def create_website_settings
-      @instructions[:settings].each { |setting| create_setting!(setting["name"], setting["value"]) }
+    def create_website_settings(website_settings_instructions)
+      website_settings_instructions.each { |setting| create_setting!(setting["name"], setting["value"]) }
     end
 
     def client_services
-      @client_services ||= ClientServices.new
+      client_services ||= ClientServices.new
```
thescubageek: I don't think this is going to work as intended, since now the variable has the same name as the method `client_services`

lib/seeder/website_seeder.rb
```diff
@@ -44,32 +46,18 @@ def create_general_settings
     end
 
     def general_settings
-      client_settings.merge(location_settings).merge(other_settings)
-    end
-
-    def client_settings
```
thescubageek: These were originally broken into the various _settings methods because of CodeClimate scores... have you checked to see what the CC effect of lumping them back together is?

variousred: no, but i would have thought it would like less methods rather than more

lib/seeder/website_seeder.rb
variousred: woops, i cought one other thing like that but missed this one. Good catch

variousred:
@thescubageek this will update the navigation settings, remapping the new page ids onto the settings hash, with page lookup by name

variousred:
@levibrown @thescubageek @quady

lib/seeder/website_seeder.rb
```diff
@@ -135,8 +124,28 @@ def has_valid_instructions?
...
-        setting.value = value
+      setting = website.settings.find_or_create_by(name: name)
+      setting.value = value
+      setting.save
+    end
+  end
+end
+
+class NavigationSettingRemapper
+  def self.remap(navigation_widget_settings, web_templates)
```
thescubageek: https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&ved=0CAcQjRxqFQoTCImE0aKen8cCFYajiAodJ0IGQg&url=http%3A%2F%2Fgrowingleaders.com%2Fblog%2Fhelp-my-brain-is-full%2F&ei=s_nIVcjsBo20ogTw2L1w&bvm=bv.99804247,d.cGU&psig=AFQjCNHwF8mIrJdkDFQFdOKaRqehe9RahQ&ust=1439320886217797

Good work on this...

PR #772, opened 2015-08-07
2.x.x theme ids

lance-hardy:
deleted hi-rise theme
reordered theme ids
renamed some themes per product instructions


thescubageek:
@lance-hardy Tests should be passing now, I had to do some clean up on the specs and the migrations..... also added three more fields as per Product's request:
* group - used to group collections of themes together (aka Berkshire)
* theme_type - location / corporate / global
* verticals - ['apartments','senior-living','self-storage']

thescubageek:
@lance-hardy @sheena-gygax Can you review? Thanks

thescubageek:
I've added the field `theme_id` to themes, which is a unique integer id per theme... this lets use rename themes (and even their directories) without breaking stuff...... Lance and I have done the work to rename a few of the themes already.... I've also added `group` (for grouping themes like Berkshires), `theme_type` (location | corporate | global ), and `verticals` (['apartments', 'senior-living', 'self-storage'])

lance-hardy:
@thescubageek just curious about  db/migrate/20150807212324_add_theme_ids.rb
The list of names are the old names?

crissmancd:
Can we rename the Wellborne themes to use 2 'L's, "Wellborne"

thescubageek:
@lance-hardy Yes they are the old names. But this brings up a good point with the PostDeployUpdaterJob that we need to consider before this rolls out...

thescubageek:
Yes @crissmancd if you look at the index.json files in for both of those themes you will see that is has been renamed to use double-L's

crissmancd:
My bad, I see it now. Bye bye Highrise. Didn't know that client moved off of it

sheena-gygax:
@thescubageek: I like it. Just out of curiosity, what will happen when then is deployed to a CMS (say like Randall or WH Long) that have a ghost template that is not tied to a theme? Those two have Wellborne Traditions V1...will they get blown away since they wont have a theme id associated to them?

thescubageek:
@sheena-gygax Yes they will. Question: why are they still on a ghost template?

thescubageek:
@lance-hardy I've added a bash script that will prevent the GardenWebThemeUpdaterJob from running until _after_ all DB migrations have finished..... this ensures that existing Themes (with their old names) are correctly migrated to the right ID _before_ we tell the GardenWebThemes to update themselves with their new names.

app/models/garden_web_theme.rb
```diff
@@ -36,5 +40,6 @@ def set_defaults
     self.javascripts ||= []
     self.layouts ||= []
     self.private_clients ||= []
+    self.verticals ||= []
```
jcstringer: Do you actually need this set_defaults logic at all? I would think that if you have serialize on the attribute with Array as the second arg, Rails would default it to an empty array for you and you could remove the set_defaults callback entirely.

db/migrate/20150807204806_add_group_type_and_vertical_to_garden_web_themes.rb
```diff
@@ -0,0 +1,13 @@
+class AddGroupTypeAndVerticalToGardenWebThemes < ActiveRecord::Migration
+  def up
+    add_column :garden_web_themes, :group, :string, :default => ""
+    add_column :garden_web_themes, :theme_type, :string, :default => "global"
+    add_column :garden_web_themes, :verticals, :text, :default => ['senior-living', 'self-storage', 'apartments']
```
jcstringer: If this attribute is supposed to serialize to an array, do you want to use an array column type instead of just text? I think you can just add array: true to this. I did some googling and couldn't really find a clear answer to my own question as far as if you gain anything by doing this (or if it even works for that matter) but I do know that newer versions of postgres have a native array column type that might perform slightly better.

app/models/garden_web_theme.rb
thescubageek: @jcstringer No I don't think we do, it was legacy from the old microformat stuff

db/migrate/20150807204806_add_group_type_and_vertical_to_garden_web_themes.rb
thescubageek: I used this method of :text when I did the widgets, I will look into array: true

thescubageek:
@lance-hardy @jcstringer This has passed QA and is being merged into v2.1.2: https://github.com/g5search/g5-content-management-system/pull/778

PR #774, opened 2015-08-07
2.0.x chad boilerplate 2

crissmancd:
Additional boilerplate mixins and classes
Update to Naked theme


public/static/components/themes/naked/layouts/index.html
```diff
@@ -0,0 +1,42 @@
+<div class="rows-layout layout">
```
thescubageek: Why are we adding a custom layout to the naked theme?

thescubageek:
@crissmancd Good refactors, just one :question: :arrow_up:

crissmancd:
@thescubageek The naked theme is now using the first boilerplate header, which needs it's own html. The header is the only difference in the layout

thescubageek:
@crissmancd This has been merged into 2.1.1-merge: https://github.com/g5search/g5-content-management-system/pull/769

PR #779, opened 2015-08-11
adds color change to text-light and dark row classes

crissmancd:
Adds color change to the row itself to account for text not within text elements, such as a div


thescubageek:
Looks good, merged into 2.1.2-merge: https://github.com/g5search/g5-content-management-system/pull/778

PR #783, opened 2015-08-12
[WIP] Update g5_header engine

bbauer:
- [x] Update g5_header (requires update of g5_updatable and g5_authenticatable)
- [x] Add centralized config
- [x] Run g5_authenticatable migrations
- [ ] Test on a staging env


db/schema.rb
```diff
@@ -109,17 +132,6 @@
   add_index "g5_updatable_locations", ["uid"], name: "index_g5_updatable_locations_on_uid", using: :btree
   add_index "g5_updatable_locations", ["urn"], name: "index_g5_updatable_locations_on_urn", using: :btree
 
-  create_table "garden_web_layouts", force: true do |t|
```
bbauer: @thescubageek were these schema removals never checked in or did my migration run improperly? Want to make sure these two tables should actually be removed...

thescubageek: It sure doesn't look like they were ever merged in. This _is_ correct, we don't need that table any more. I will update 2.1.2-merge so that it contains the correct schema

bbauer:
Closing in favor of https://github.com/g5search/g5-content-management-system/pull/785

PR #785, opened 2015-08-12
[WIP] Update g5_header engine (2.1.2-merge target)

bbauer:
This is a new PR of https://github.com/g5search/g5-content-management-system/pull/783 which changes the target branch to 2.1.2-merge from master.

- [x] Update g5_header (requires update of g5_updatable and g5_authenticatable)
- [x] Add centralized config
- [x] Run g5_authenticatable migrations
- [x] Test on a staging env


thescubageek:
@bbauer I undid the upgrade to the latest g5_header on 2.1.2 since the CSS was breaking the CMS when precompiled. Let's see what we can do to namespace the CSS in g5_header such that it doesn't bleed into other apps.

PR #789, opened 2015-08-12
sets the heroku app tier to hobby after creation,

variousred:
to prevent sleep


thescubageek:
Big effing win :rainbow: I'll merge it in: https://github.com/g5search/g5-content-management-system/pull/778

PR #791, opened 2015-08-13
WIP: CMS v2.1.3 Merge

thescubageek:
### CMS Architecture:
* Adds optional custom slugs to single domain location URLs
* Updates G5Header to use latest version, which includes centralized apps
* Adds `rake resque:kill_all` and `rake resque:kill[id]` rake tasks for clearing out stuck Resque workers
* Adds `rake resque:all` and `rake resque:working` for getting a list of IDs for Resque workers
* Updates all Location Simple Spinup files to use latest SEO strategies

### Widgets:
* Calls To Action: Adds heading and heading position settings
* Contact Info: Adds social icons and links
* Promoted Reviews: Adds an optional 'Read More' link

### Themes:
* CFLane: Makes minor font size CSS tweaks
* Holland Corp: Adds custom styling to the Multifamily Search widgets

PR #794, opened 2015-08-14
adds kill resque workers rake task

thescubageek:
Tired of your workers being stuck and lazy? Kill 'em all and let god sort 'em out.


lib/tasks/resque.rake
```diff
@@ -6,3 +6,8 @@ end
  
 desc "Alias for resque:work (To run workers on Heroku)"
 task "jobs:work" => "resque:work"
+
+desc "kills all Resque workers"
```
steveburkett: Hey Steve.  Just 2 cents here:  I had do deal with this in Reputation a few times.   What i did was this:

Resque.queues.each{|q| Resque.redis.del "queue:#{q}" }
Resque.workers.each {|w| w.unregister_worker}

To reset Resque operationally.  The workers would then be unregistered and queues deleted, so that resque could recreate the queues the next time.   

Last, i had to go into bash and find the orphaned resque workers and kill -2 them (this part can be shell script automated, i never got around to doing this, but these pids will hang around forever if not taken care of.

lib/tasks/resque.rake
```diff
@@ -6,3 +6,8 @@ end
...
+task "resque:kill" => :environment do
+  Resque.workers.each { |w| w.unregister_worker }
```
davidw: Is there any way that unregister_worker can fail?

lib/tasks/resque.rake
davidw: This looks useful: https://gist.github.com/jeanmartin/992680

http://www.theallium.com/engineering/computer-programming-to-be-officially-renamed-googling-stackoverflow/ :-)

thescubageek: I think a combo of what @steveburkett suggested above and the guys in the gist ^^ does with his `rake resque:stop_workers` (where he records the pids and then kills them) would do a solid job

lib/tasks/resque.rake
thescubageek: Not that I'm aware of...

thescubageek:
Made changes to the rake tasks to do the following:

* `rake resque:kill_all` - kills all resque workers and clears queues
* `rake resque:kill[id]` - kills one resque worker matching id
* `rake resque:working` - finds ids for all working resque workers
* `rake resque:all` - finds ids for all resque workers

thescubageek:
merged into https://github.com/g5search/g5-content-management-system/pull/791

PR #796, opened 2015-08-17
Ivy BP Map Issues

aborows:
https://www.pivotaltracker.com/n/projects/1282248/stories/100855826


public/static/components/themes/ivy/stylesheets/ivy.scss
```diff
@@ -1205,12 +1205,14 @@ input[type=submit] {
   .text {
     width: 100%;
   }
+  .text{
```
thescubageek: @aborows I'll need a little explanation as to what's going on here.... 8em is a very big height to have hardcoded...

aborows: @thescubageek 8em was a bit excessive. I switched it to 7em. I am replacing that top:7em; applied to .canvas with height:7em; to .text, so the map and directions do not overlap.

screen shot 2015-08-17 at 2 45 25 pm

to...

screen shot 2015-08-17 at 2 43 49 pm

thescubageek: My bigger concern is that the use of a hard-coded height isn't the right solution..... the directions DIV should be using position: relative, display: block and clearfix to get put in the correct layout.... using a hardcoded height is smelly to me, seems like it's a bandaid for something that is actually a hasLayout issue. I saw this similar "fix" in core a lot and it caused issues down the road when doing responsive.... for instance, 7em on mobile will be massive amount of space.

aborows: Steve, you're right. This is more of a bandaid to fix the required items. I'll redo this for Wednesday, so it is done correctly. Thanks!

PR #800, opened 2015-08-17
2.1.3 holland corp

lance-hardy:
- removed padding from .col for multifamily search result page only
- set margin between two search buttons at odd media range for multifamily search results page


thescubageek:
Merged https://github.com/g5search/g5-content-management-system/pull/791

PR #801, opened 2015-08-17
Redirects 100227692

variousred:
User Story
"As a G5 user, I want the ability to place redirects in the system with error handling, so that we can ensure that client sites have redirects that work."

Acceptance Criteria

    Add regex validation on redirect input fields, so that if a user enters a leading or trailing slash, they are given a system error
    Add confirmation message to redirects manager 

https://github.com/G5/orion-cms/pull/45
https://www.pivotaltracker.com/story/show/100227692


app/models/web_template.rb
```diff
@@ -192,6 +194,21 @@ def application_min_css_path
...
+      validate_redirect_path(redirect_path)
+    end if redirect_patterns.present?
+  end
+
+  def validate_redirect_path(redirect_path)
+    case
+    when redirect_path.include?('http')
+      errors.add(:redirect_patterns, "can't be full url")
+    when !redirect_path.match(/^[^\/]+[\S]+[^\/]+$/)
+      errors.add(:redirect_patterns, "cant start or end in a slash `/` `#{redirect_path}`")
```
thescubageek: Looks like with this rule than //mydomain.com is invalid as well, is that correct?

thescubageek:
Update the controller specs to pass, and I think we're good here as long as the redirect rule I mentioned ^^ is intended...

PR #803, opened 2015-08-18
CMS v2.1.4

thescubageek:
### CMS Architecture:
* Adds regex validation on redirect input fields, so that if a user enters a leading or trailing slash they are given a system error
* Adds confirmation message to redirects manager

### Widgets:
* Locations Navigation: Changes CSS overrides to use class-level specificity instead of ID-level (makes it easier for themes to override colors if necessary)
* Converts the following widgets to use the new WidgetEditorFilter: News Feed, Promoted Reviews, Realpage, Social Feed, Walkscore, Video
* Promoted Reviews: Adds an optional 'Read More' link (for real the time!)

### Themes:
* Fusion: Changes Locations Navigation widget to match style of Calls To Action widget in Hero Content Stripe
* Ivy: Patches overlap bug on Directions widget
* Luxe: Tweaks Locations Navigation widget to use secondary color instead of tertiary color


thescubageek:
:watch: 2 :ship:

PR #806, opened 2015-08-18
2.1.X Click-2-Edit

thescubageek:
* Adds new Page Editor (a _much_ needed overhaul)
* Adds Click2Edit widget buttons in preview

This branch has been moved to: https://github.com/g5search/g5-content-management-system/pull/869


app/assets/javascripts/web_template/click_2_edit.js
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  Click2Edit.prototype.insertErrorMessages = function(errors) {
+    var error = "<div class='alert alert-error'>" + errors["errors"]["base"][0] + "</div>";
+    return $('#modal .modal-body', this.parentDoc).prepend(error);
+  };
+
+  return Click2Edit;
+
+})();
```
jcstringer: Seems painful to have to write this much raw js when coffeescript can make it so much more concise! Def not saying you should go back and rewrite it all, but I do think it highlights the fact that this particular app has a big mix of js and coffee in the assets as well as the widgets. As a team we should probably standardize on something in the long run across all our apps. I'd love it to be coffeescript and I'm sure @levibrown would love it to be es6.

levibrown: yep, es6 is the way to go imo. we are actively migrating the cms ember side to es6, should do the same here.

also, not a fan of all the prototypes above... you could add all they click2edit methods in a object to make it more readable.
Something like this?:
```
var ClickToEdit = {
    widgetExceptionList: function() {
      return {
        "photo": "img",
        "contact-info-sheet": ".info-sheet-nav"
    };
  },
  ...
}
```

thescubageek: Story on this one is that I started on CoffeeScript, got it 85% of the way, hit a snag with the way that coffee was compiled in the asset pipeline, converted it to JS using js2.coffee, and then just rode their interpretation of it the rest of the way (hence the prototype ugliness).

I think we should go for ES6 with this upgrade... take this chance to go next-gen and clean this up. What gem do you reckon I should use...? It has to be compatible with the precompile... I was looking at [Babel](https://babeljs.io/docs/setup/#rails) but noticed this warning: https://github.com/TannerRogalsky/sprockets-es6#asset-manifests-required-for-precompiling

app/assets/javascripts/web_template/click_2_edit.js
```diff
@@ -0,0 +1,415 @@
+if (typeof(window.noEvent) === 'undefined') {
```
thescubageek: @jcstringer @levibrown Got rid of all those nasty .prototypes

thescubageek:
Merge with master didn't go so well, so I extracted the important changes into a new branch: https://github.com/g5search/g5-content-management-system/pull/869

PR #807, opened 2015-08-19
grab the newest version of g5_heroku_app_name_formatter

johnlucia:
There's a new g5-geo-service that will take care of the locations_search requests coming from live sites. The Hub has been serving those requests to this point which isn't super cool. 

The new version of g5_heroku_app_name_formatter references that new centralized app


johnlucia:
Apparently you can get worms by the barrel. Who knew?

PR #811, opened 2015-08-19
CMS v2.1.5 Merge

thescubageek:
### CMS Architecture:
* Updates `g5_header` gem to use latest version
* Speeds up and reduces memory requirements for `GardenWidgetUpdaterJob`
* Patches sites to use `<meta charset="utf-8">`, not the typo-ridden `<meta charset="utf8">`
* Updates CMS development documentation on the CMS Wiki

### Widgets:
* Navigation: Triggers `navigationReady` event after Navigation widget is set up

### Themes:
* Fusion: Patches Navigation height bug on desktop during page load
* Ivy: Patches overlap bug and canvas width bug on Directions widget
* Luxe: Patches overflow bug and hides border on Hero image

PR #817, opened 2015-08-23
Fusion Menu Fix

aborows:
https://www.pivotaltracker.com/n/projects/1282248/stories/101636708


public/static/components/themes/luxe-version-b/javascripts/luxe.js
```diff
@@ -4,9 +4,14 @@
...
-      $this.addClass('stay-put').wrap('<div class="stay-put-wrapper"></div>');
-      $this.parent().css('height', wrapHeight);
+      var checkExist = setInterval(function() {
+        if ($('#drop-target-nav .navigation').length) {
+          var wrapHeight = $this.outerHeight();
+          $this.addClass('stay-put').wrap('<div class="stay-put-wrapper"></div>');
+          $this.parent().css('height', wrapHeight);
+          clearInterval(checkExist);
+        }
+      }, 100); // check every 100ms
```
thescubageek: `SetIntervals` are _usually_ a smell-test of something else that could be done better.... in this case, we're going to have 100ms polling until the nav is part of the DOM.... why?

thescubageek:
Looks good after we added the `navigationReady` event to the Navigation widget... that way all Themes can use this in the future. Merged into: https://github.com/g5search/g5-content-management-system/pull/811

PR #821, opened 2015-08-24
Updates to MF IUI cards to support realpage skip step

jcstringer:
https://www.pivotaltracker.com/story/show/100072836


public/static/components/widgets/multifamily-iui-cards/index.json
```diff
@@ -46,11 +51,34 @@
         "default_value": ""
       },
       {
+        "name": "secondary_success_redirect_url",
+        "editable": true,
+        "default_value": ""
+      },
+      {
+        "name": "search_url",
```
thescubageek: `search_url` and `site_id` should have empty `default_value` attributes

public/static/components/widgets/multifamily-iui-cards/edit.html
```diff
@@ -12,9 +12,21 @@
     />
```
thescubageek: Let's pair up to convert this over to liquid edit and get it quickly tested out.

thescubageek: I've converted this over to use the new liquidEdit.... but I think we need to clarify a few of these variables so it's easier to use.

thescubageek:
Let's make sure the new edit form works after converting it to liquid..... then let's let Byron know of the changes and we should be good to go.

jcstringer:
Closing out this PR as this branch is going to get merged into a later release.

PR #826, opened 2015-08-25
In a little hilltop village, they gambled for my clothes

johnlucia:
- Using session storage for PPC numbers causes numbers to stick when navigating between different locations under the same single domain client. Fix that by just looking up the number in the Json.

- Tweak city-state-zip-search to use a new soon-to-be-added `home_page_url` attribute in the location_search json response from the hub.


public/javascripts/libs/phone-service.coffee
```diff
@@ -159,7 +162,7 @@ class NumberFinder
     newCpmCode = @getParameterByName('cpm')
     cpmCode = newCpmCode || sessionStorage.getItem('cpm')
 
-    ppcNumber = sessionStorage.getItem(cpmCode) if sessionStorage.getItem(cpmCode) 
+    ppcNumber = @loc["#{cpmCode}"] if sessionStorage.getItem(cpmCode)
```
jcstringer: Not sure but you may be able to just use 
```
@loc[cpmCode] if sessionStorage.getItem(cpmCode)
```
since the key is assumed to be a string

jcstringer:
:ship:

johnlucia:
merged this here instead...  https://github.com/g5search/g5-content-management-system/commit/776c2aa51b68a9961124feec07c0a18e60050ba6

PR #827, opened 2015-08-25
Cobalt product warranty round one fixes, html boilerplate change

lance-hardy:
- add new data images
- fix font weight issue for FF browser
- change heading lines schema
- rework subnav icon - replace content with data image
- restore widget headings
- mod quote presentation
- mod city state zip search widget styles
- separate multifamily search styles
- mod html widget default boilerplate styles


thescubageek:
Only reason your spec was failing is because of a known bug in our system when Faker uses apostrophized names, eg:
```
  Failure/Error: expect(@json['yml']).to eq(file_name)
       
       expected: "apartments_kennith_weber_kyleigh_o_connell"
            got: "apartments_kennith_weber_kyleigh_o'connell"
       
       (compared using ==)
```

This is fixed by rerunning the spec...

PR #832, opened 2015-08-26
Asset seeder serializer

variousred:
local asset references were not being restored with a save restore. this serializes and seeds these records


thescubageek:
@mmitchellg5 This has been merged into: https://github.com/g5search/g5-content-management-system/pull/825

PR #833, opened 2015-08-26
remove microformats gem

thescubageek:
We don't use it, so why keep it? :nuke:


johnlucia:
:+1:

jcstringer:
+1 this is great. There are a couple of misc spots where microformats are still mentioned which should prob get cleaned up. Just comments and a README so I'd say they can just be changed with a PR

https://github.com/g5search/g5-content-management-system/search?utf8=%E2%9C%93&q=microformats

PR #834, opened 2015-08-26
2.1.6 query optimize

thescubageek:
* Speeds up a few things in WidgetDrop by shortcutting the parent Settings lookup (Widget model handles this instead)


johnlucia:
:+1:

PR #838, opened 2015-08-26
Self storage search home chad

crissmancd:
WIP


crissmancd:
@dreuss please take a look. let me know what questions you have and if i royally screwed anything up. Everything appears to look and function the same.

public/static/components/widgets/self-storage-search-home/show.html
```diff
@@ -10,51 +10,50 @@
...
 
-  <div class="sss-search-criteria">
+  <form class="sss-search-criteria" action="">
     <h2>Find a Storage Unit</h2>
-    <label for="search-input" class="search-input">
-      <span class="twitter-typeahead">
-        <input class="sss-search-input typeahead tt-hint twitter-typeahead" type="text" placeholder="Search by City or State">
-      </span>
+    <div class="search-input">
+      <input class="sss-search-input typeahead tt-hint" type="text" placeholder="Search by City or State">
```
dreuss: would like to keep the twitter-typeahead class in here.  may not be referenced in the scss, but could be a dependency of the typeahead.js library...just in case.

public/static/components/widgets/self-storage-search-home/stylesheets/self-storage-search-home.scss
```diff
@@ -1,200 +1,188 @@
...
     background-color: $color3;
-    // color: $color4;
     padding: 23px 23px 38px;
     width: 42%;
-    float: none;
-    overflow:visible;
+    overflow: visible;
     margin: 0 auto;
-    h2{
-      letter-spacing:0;
```
dreuss: Some themes implement some wide letter spacing on h2's etc, so this is a combative measure to not jack the styles in the widget.

public/static/components/widgets/self-storage-search-home/show.html
crissmancd: I noticed that the plugin adds the span with "twitter-typeahead" class automatically, which is why i removed it (was creating 2 spans with that class). also all of the docs and examples of the plugin only include "typeahead" class on the input. If we keep the twitter-typeahead class here then the span and input will be getting the same styles.

Maybe I'm missing something?

dreuss:
Good due diligence here, appreciate the extra set of eyes.  Am going to merge this as the final stamp is still awaiting, and we'll see this adversely affects it's state right now.  Thanks.

PR #840, opened 2015-08-26
Fusion Menu Fix

aborows:


public/static/components/themes/luxe-version-b/javascripts/luxe.js
```diff
@@ -4,12 +4,20 @@
 
     init: function() {
       var $this = $(this);
-      $(window).on('navigationReady', function() {
+      if ($('#drop-target-btn > span').length > 0) {
         var wrapHeight = $this.outerHeight();
```
crissmancd: Can you wrap the 3 lines in this if statement into a function and use it down below too. Less code duplication is good. then we only need to make a change in one place

crissmancd:
After the one fix it's good to go :rainbow:

thescubageek:
Shipping it without the refactor, since the refactor was breaking stuff and we need this patch out soon. https://github.com/g5search/g5-content-management-system/pull/825

PR #841, opened 2015-08-26
2.1.4ish cobalt qa v1

lance-hardy:
last round of QA requests. 
- change margin % for html widget
- tighten zip search widget styles (in theme)


public/static/components/widgets/html/stylesheets/html.css
```diff
@@ -16,14 +16,16 @@
...
 
-.row-single .html {
-  text-align: center;
-}
-
 .row-single .html-content {
   margin-left: auto;
   margin-right: auto;
-  max-width: 50em;
+  max-width: 75%;
```
thescubageek: @lance-hardy I'm a little worried about this line since it will affect _every_ HTML widget across all themes. Can you show me an example of what this fixed... and how can we confirm that it hasn't broken other themes?

lance-hardy: @thescubageek this addresses new styles added to the html widget not long ago that constrained max width of the widget to 80em and assigned max width 50em to the content class. Christie asked to change this as the html content in a single col row was too 'narrow' as a result. Changing to percentages allows for a more consistent max width and margin regardless of site width.

thescubageek: @lance-hardy After conversation with Christie T, it makes sense to get the changes out now. :ship:

PR #846, opened 2015-08-27
2.1.7 cms url

thescubageek:
* Refactors Formatter and CMS url methods into separate concern, which fixes bug in GardenWebThemeUpdater.
* Patches Naked Theme


app/models/widget.rb
```diff
@@ -4,6 +4,7 @@ class Widget < ActiveRecord::Base
   include WidgetSettings
   include UpdateNavSettings
   include LiquidParameters
+  include CmsUrlFormatter
```
thescubageek: This is necessary for future work on Click2Edit

thescubageek:
@crissmancd @jcstringer Quick review on this one? It gets custom layouts to work on Naked plus refactors some ugly code

app/models/cms_url_formatter.rb
```diff
@@ -0,0 +1,15 @@
+module CmsUrlFormatter
```
jcstringer: I think you prob meant to put this file in the concerns folder instead of models

jcstringer:
Nice work DRY'ing up all that code!

app/models/cms_url_formatter.rb
thescubageek: D'oh, of course I did. I had this set up correctly on Click2Edit branch but needed it in earlier to make the theme fix work, moved it over all wrong...

PR #855, opened 2015-08-28
2.1.7 rake restart

thescubageek:
* Adds `rake deploy:restart` task so Heroku Scheduler can clear up R14s every hour


maeve:
:+1:

PR #858, opened 2015-08-28
Geo service

johnlucia:
There's a new service dedicated to location searches so the hub isn't directly serving those requests to live sites.

This points all new implementations to that service, and creates a rake task so we can methodically migrate the existing widgets over to the new service...  Or batch change any other setting in the future.

Not jumping on the liquid bandwagon for the widget setting just yet, since I'd like to test the new service with just a couple clients to start with.


thescubageek:
Looks good, :+1: on the new rake task, I had to manually do something to that effect earlier via command line and it was a pain.

thescubageek:
When is the new geo-service ready?

johnlucia:
Right about meow...

Old Hub Json:

https://g5-hub.herokuapp.com/clients/g5-c-1t2t99el-urban-self-storage/location_search.json?search=tacoma%20wa&radius=20

New g5-geo-service Json:
http://g5-geo-service.herokuapp.com/clients/g5-c-1t2t99el-urban-self-storage/location_search.json?search=tacoma%20wa&radius=20

PR #860, opened 2015-08-28
WIP: 2.x.x config endpoint

thescubageek:
* Migrates web_template and parent_widget on Widget model to use field on model, not convoluted settings lookups -- should speed up performance.
* Updates config endpoint to be much more memory efficient


thescubageek:
Gonna reopen this one against master

PR #867, opened 2015-08-31
CMS v2.1.8 Merge

thescubageek:
### CMS Architecture:
* Fixes `generic_text_field` to have placeholder text if the edit field is not utilized for liquid values
* Patches Saves to restore Content Stripe background and text color settings
* Fixes missing trailing slash bug on single domain Location urls
* Adds `activerecord-session_store` gem, which uses local sessions rather than cookies for storage

### Widgets:
* Fixes bug where social media links break when Contact Info and Footer Info widgets co-exist
* Fixes bug where CSS "bleed-in" was occurring when Featured Properties and Gallery widgets co-exist
* Sitelink eLease IUI Cards: Gracefully handles situations where no units are present for a Location

### Themes:
* CFLane: Removes pending status and makes minor link color tweaks


thescubageek:
Featured Properties and Gallery fix has been reverted until @crissmancd and the theme team have time to patch up the theme overrides to work with this change.

public/static/components/widgets/gallery/show.html
```diff
@@ -1,4 +1,4 @@
-<div class="gallery widget">
+<div class="gallery gallery-only widget">
```
variousred: could this be more specific? or make the competing one that clashes more specific

crissmancd: Steve and I have a fix for this going out today. We namespaced this and the featured properties widgets better

Chad

Sent from my iPhone

> On Sep 2, 2015, at 6:20 PM, Michael Mitchell <notifications@github.com> wrote:
> 
> In public/static/components/widgets/gallery/show.html:
> 
> > @@ -1,4 +1,4 @@
> > -<div class="gallery widget">
> > +<div class="gallery gallery-only widget">
> could this be more specific? or make the competing one that clashes more specific
> 
> —
> Reply to this email directly or view it on GitHub.
> 

-- 
*This email may contain information that is privileged, confidential, or 
proprietary and is intended solely for the named addressee. If you are not 
the addressee, or if it appears that you have received this email in error, 
please advise me immediately by reply email, do not disclose, copy, or 
distribute the contents, and immediately delete the message and any 
attachments from your system. Thank you.*

PR #872, opened 2015-09-01
Generic text placeholder

dreuss:
This fixes the case when using placeholder text that isn't liquid.  By default the placeholder would default to a blank string, which needed to be some string value.  Now the helpers edit_text_field or edit_color_field should use the default configs correctly.


thescubageek:
:+1: Merged into 2.1.8-merge: https://github.com/g5search/g5-content-management-system/pull/867

PR #874, opened 2015-09-01
For a time I would feel I belonged still to a world of straightforwar…

johnlucia:
Fix the js in the footer-info and contact-info widgets so they don't edit a bunch of attributes twice when they coexist on a page, and subsequently spray a bunch of 404 errors into the console and break SVG's in firefox.

also had to undo a direct edit to a JS file that needed to go into coffeescript in order to make it all work out.


thescubageek:
:+1: nice work, sorry you had to bust out the voodoo. :ship: to the merge branch

PR #875, opened 2015-09-01
2.1.8 featured loc fix

thescubageek:
* Fixes Featured Properties and Gallery widgets to be CSS namespaced so they don't bleed into each other on deployed sites


variousred:
:shipit:

variousred:
although, @thescubageek ```gallery-only``` implies that the other ISNT a gallery, which it does appear to be especially since it also has classes of ```.gallery```

PR #881, opened 2015-09-02
add sessions store and migration

thescubageek:
Adds session store gem and migration


config/initializers/session_store.rb
```diff
@@ -1,6 +1,6 @@
 # Be sure to restart your server when you modify this file.
 
-G5CMS::Application.config.session_store :cookie_store, key: '_g5-cms_session'
+Rails.application.config.session_store :active_record_store, :key => '_g5-dashboard_session'
```
steveburkett: this should be _g5-cms_session

db/schema.rb
```diff
@@ -220,6 +234,30 @@
 
   add_index "locations", ["urn"], name: "index_locations_on_urn", using: :btree
 
+  create_table "page_configs", force: true do |t|
```
steveburkett: did this come from something else?

thescubageek: Yes and I don't want that in there yet, thanks!

PR #882, opened 2015-09-02
Remove all traces WidgetEntry class, which is unused

jcstringer:
@thescubageek and I noticed that this class and its related logic and code are unused at this point. Time to clear out some cruft.


lib/hentryable_dates.rb
```diff
@@ -1,17 +0,0 @@
-module HentryableDates
-  def created_at_computer_readable
-    model.created_at.utc.to_s(:computer)
-  end
-
-  def created_at_human_readable
```
thescubageek: Human readable my ass

thescubageek:
:+1: Merged it into 2.1.9-merge: https://github.com/g5search/g5-content-management-system/pull/887

Great work gutting this!

jcstringer:
Thanks @thescubageek. We'll def want to make sure QA does full regression tests to make sure this stuff wasn't used anyplace. Pretty sure it isn't but still you never know.

PR #884, opened 2015-09-02
2.1.8 trailing slash home sd

thescubageek:
* Adds trailing slash to cannonical url for single domain location home pages


app/models/web_template.rb
```diff
@@ -145,7 +145,7 @@ def page_url
     if single_domain_internal_page?
       File.join(domain_for_type.to_s, website.single_domain_location_path.to_s, slug)
     elsif single_domain_location_home_page?
-      File.join(domain_for_type.to_s, website.single_domain_location_path.to_s)
+      File.join(domain_for_type.to_s, website.single_domain_location_path.to_s, '/')
```
variousred: whats the difference between ```url``` and ```page_url``` ?

variousred:
:shipit:

app/models/web_template.rb
thescubageek: page_url is the full url: `https://www.qatestbeds.com/senior-living/or/bend/luxe/`
url is the path: `/senior-living/or/bend/luxe`

...because it's Orion, so Simplify, right? We should rename url to be something more meaningful (url_path) in another PR

PR #886, opened 2015-09-02
Fix - Super HOT, Really HOT

dreuss:
Just coming hot, rubber side down, fixin category select boxes.


thescubageek:
:fire: :fire_engine: .... :ship:

PR #889, opened 2015-09-03
Change widget defaults and name for Self Storage eLease IUI

jcstringer:
https://www.pivotaltracker.com/story/show/101920432


public/static/components/widgets/self-storage-iui-cards-v2/index.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "Sitelink eLease IUI Cards",
+  "name": "Self Storage Elease IUI Cards",
```
thescubageek: Really long string, how does this label look in the widgets panel of the CMS?

jcstringer: Seems ok to me

screen shot 2015-09-03 at 2 40 19 pm

PR #890, opened 2015-09-03
Heart of darkness

johnlucia:
Fixes a bug in city level area pages. 

Everything is fine if ALL locations within a given city are assigned a neighborhood value, or NO locations within a given city are assigned a neighborhood.  If some have neighborhoods and some don't, The CMS generates an area page for a neighborhood of `""`, which in a round-about way, breaks the city level area page by only showing the locations without a neighborhood value.

Kind of hard to explain, so you can back out the little change in lib/static_website/compiler/area_pages.rb and run the new spec, and you'll get the idea.


lib/static_website/compiler/area_pages.rb
```diff
@@ -58,7 +58,7 @@ def cities_for_state(state)
       end
 
       def neighborhoods_for_city(city)
-        websites_for_city(city).map { |website| website.owner.neighborhood_slug }.uniq
+        websites_for_city(city).map { |website| website.owner.neighborhood_slug }.uniq - [""]
```
variousred: I'd prefer to take care of this higher up the chain with not saving empty strings into the DB to begin with, maybe something like proposed here http://stackoverflow.com/questions/7202319/rails-force-empty-string-to-null-in-the-database

that said, this does fix the issue at hand

johnlucia: I'm not sure if can use that since all this is being pulled out of Location.properties which is that json datatype, so we don't get to use a lot of magic that would normally be available in a rails app. 

...I'm not entirely sure about that statement since my understanding of the way that's working in here is foggy at best

variousred: Ugh, yeah, you're right

variousred: If this were fixed at the hub and neighborhoods were nil, it looks like they would serialize as nil as well according to what i can tell from this: http://api.rubyonrails.org/classes/ActiveModel/Serialization.html

johnlucia:
closing this PR since this is getting merged into the release branch... https://github.com/g5search/g5-content-management-system/pull/892

PR #891, opened 2015-09-03
serializes content stripe text and bg colors

variousred:


thescubageek:
Merging it in as a hotfix

